PR #192, opened 2023-03-08
Add dependency injection for core SQRL frontend

henneberger:


sqrl-planner/src/test/java/com/datasqrl/plan/local/analyze/RetailSqrlModule.java
```diff
@@ -0,0 +1,358 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+public class RetailSqrlModule implements SqrlModule {
```
mbroecheler, line 52: This class looks strange to me. There is so much code and type mapping in here. Is that really necessary? Is it not possible to use one of the type converters for this?

sqrl-planner/src/test/java/com/datasqrl/plan/local/analyze/RetailSqrlModule.java
```diff
@@ -0,0 +1,358 @@
...
+            ))
+        ));
+      } else {
+        throw new RuntimeException();
+      }
+    }
+    return new RelationType<>(fields);
+  }
+
+  private Type toType(Column type) {
```
mbroecheler, line 166: Should this really live here?

Issue #231, opened 2023-04-14
Implement asset compilation


henneberger:
The goal of this ticket is to implement a modular and extensible architecture for artifact generation, asset compilation, and (later) deployment. The solution should support multiple environments, build targets, and deployment assets, such as Flink JARs and HTTP server executables.

Proposed Solution:
1. (prereq) Add http server as an 'engine' target.
2. Create a Java interface for asset compilation, which engines can extend and implement.
3. Use ServiceLoader to load and invoke the appropriate asset compilation implementation based on the user configuration.
4. Implement asset compilation using TestContainers to run Docker containers for each build environment, allowing asset generation to happen within the same JVM process.

PR #238, opened 2023-04-26
Add servlets as engine types (#231)

henneberger:
Adds servlets as a 'server' engine type. Currently it is a no-op for planning since the engine type currently has no local query engine. This is a necessary prerequisite for adding the ability to generate deployment assets.

### DIFF
--- a/sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/config/PipelineFactory.java
+++ b/sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/config/PipelineFactory.java
@@ -87,11 +87,9 @@ public class PipelineFactory {
 
   public Optional<ServerEngine> getServerEngine() {
     Collection<ExecutionEngine> engines = getEngines(Optional.of(Type.SERVER)).values();
-    if (engines.isEmpty()) {
-      return Optional.empty();
-    }
 
-    return Optional.of((ServerEngine)engines.stream().findFirst().get());
+    return engines.stream().findFirst()
+        .map(e->(ServerEngine) e);
   }
 
   public ExecutionPipeline createPipeline() {
--- a/sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/config/PipelineFactory.java
+++ b/sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/config/PipelineFactory.java
@@ -10,6 +10,7 @@ import com.datasqrl.engine.database.DatabaseEngine;
 import com.datasqrl.engine.database.DatabaseEngineFactory;
 import com.datasqrl.engine.pipeline.SimplePipeline;
 import com.datasqrl.engine.pipeline.ExecutionPipeline;
+import com.datasqrl.engine.server.ServerEngine;
 import com.datasqrl.engine.stream.StreamEngine;
 import com.datasqrl.metadata.MetadataStoreProvider;
 import java.util.Collection;
@@ -84,12 +85,19 @@ public class PipelineFactory {
     return (StreamEngine) getEngine(Type.STREAM);
   }
 
+  public Optional<ServerEngine> getServerEngine() {
+    Collection<ExecutionEngine> engines = getEngines(Optional.of(Type.SERVER)).values();
+    if (engines.isEmpty()) {
+      return Optional.empty();
+    }
+
+    return Optional.of((ServerEngine)engines.stream().findFirst().get());
+  }
+
   public ExecutionPipeline createPipeline() {
     DatabaseEngine db = getDatabaseEngine();
     StreamEngine stream = getStreamEngine();
-    return SimplePipeline.of(stream, db);
+    Optional<ServerEngine> server = getServerEngine();
+    return SimplePipeline.of(stream, db, server);
   }
-
-
-
 }
--- a/sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/engine/EngineCapability.java
+++ b/sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/engine/EngineCapability.java
@@ -34,7 +34,9 @@ public enum EngineCapability {
   //Writing/upserting data into engine by primary key will deduplicate data
   MATERIALIZE_ON_KEY,
   //Engine supports data monitoring
-  DATA_MONITORING;
+  DATA_MONITORING,
+  //No capabilities
+  NONE;
 
 
   public static EnumSet<EngineCapability> STANDARD_STREAM = EnumSet.of(DENORMALIZE,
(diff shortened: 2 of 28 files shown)


sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/config/PipelineFactory.java
```diff
@@ -84,12 +85,19 @@ public StreamEngine getStreamEngine() {
     return (StreamEngine) getEngine(Type.STREAM);
   }
 
+  public Optional<ServerEngine> getServerEngine() {
+    Collection<ExecutionEngine> engines = getEngines(Optional.of(Type.SERVER)).values();
+    if (engines.isEmpty()) {
+      return Optional.empty();
+    }
+
+    return Optional.of((ServerEngine)engines.stream().findFirst().get());
```
mbroecheler, line 94: simplify to (ServerEngine)engines.stream().findFirst()

mbroecheler: you don't need the if clause either.

sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/engine/EngineCapability.java
```diff
@@ -34,7 +34,9 @@ public enum EngineCapability {
   //Writing/upserting data into engine by primary key will deduplicate data
   MATERIALIZE_ON_KEY,
   //Engine supports data monitoring
-  DATA_MONITORING;
+  DATA_MONITORING,
+  //No capabilities
+  NONE;
```
mbroecheler, line 39: Why do you need that? I am not a big fan of having a capbility NONE - that's confusing.

sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/engine/EngineCapability.java
```diff
@@ -44,5 +46,7 @@ public enum EngineCapability {
   public static EnumSet<EngineCapability> STANDARD_DATABASE = EnumSet.of(NOW, GLOBAL_SORT, MATERIALIZE_ON_KEY,
       MULTI_RANK);
 
+  public static EnumSet<EngineCapability> NO_CAPABILITIES = EnumSet.of(NONE);
```
mbroecheler, line 49: Why not just EnumSet.noneOf(EngineCapability.class)?

sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/engine/server/ServerEngine.java
```diff
@@ -0,0 +1,20 @@
...
+import java.io.Closeable;
+
+/**
+ * The server engine is a combination of the server core (the graphql engine) and the
+ * servlet that is running it. Some servlets may not support things like Java, reflection,
+ * code generation executors, etc.
+ */
+public interface ServerEngine extends ExecutionEngine {
+
+  default DataMonitor createDataMonitor() {
```
mbroecheler, line 17: What is a data monitor for a Server? Can we find a different name for that since we use data monitoring currently in the context of generating data statistics in discovery. You could also rename the latter if data monitor is a better fit for what you are trying to accomplish here.

sqrl-engines/sqrl-engine-server/src/main/java/com/datasqrl/engine/server/LambdaJavaEngineFactory.java
```diff
@@ -0,0 +1,31 @@
...
+import com.datasqrl.engine.EngineFactory;
+import com.datasqrl.engine.ExecutionEngine;
+import com.datasqrl.engine.server.LambdaNativeArm64EngineFactory.LambdaNativeArm64Engine;
+import com.google.auto.service.AutoService;
+import lombok.NonNull;
+
+@AutoService(EngineFactory.class)
+public class LambdaJavaEngineFactory extends GraphqlServerEngineFactory {
+
+  public static final String ENGINE_NAME = "aws-lambda-native";
```
mbroecheler, line 13: You have an engine for arm and x86 - what is this generic native lambda engine for?

sqrl-planner/sqrl-planner-local/src/main/java/com/datasqrl/plan/global/DAGAssembler.java
```diff
@@ -172,10 +174,20 @@ public PhysicalDAGPlan assemble(SqrlDAG dag, Set<URL> jars, Map<String, UserDefi
         });
 
     PhysicalDAGPlan.StagePlan streamPlan = new PhysicalDAGPlan.StagePlan(streamStage, writeDAG,
-        null, jars, udfs);
+        null, jars, udfs, null);
 
-    return new PhysicalDAGPlan(ListUtils.union(List.of(streamPlan),databasePlans));
 
+    Optional<ExecutionStage> serverStage = pipeline.getStage(Type.SERVER);
```
mbroecheler, line 180: I think this will look cleaner if you rewrite it to create an ArrayList and then add all the plans instead of doing a bunch of unions.

mbroecheler (changes requested):
some minor change requests and a few questions.

sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/engine/EngineCapability.java
henneberger:
```
@NonNull EnumSet<EngineCapability> capabilities
```
It's not null and an empty EnumSet is not allowed.

henneberger: Ah, maybe there is one, let me check.

sqrl-engines/sqrl-engine-core/src/main/java/com/datasqrl/engine/server/ServerEngine.java
henneberger: Ah sorry, I thought this was a needed interface.

sqrl-engines/sqrl-engine-server/src/main/java/com/datasqrl/engine/server/LambdaJavaEngineFactory.java
henneberger: Java. So if we put in calcite which does codegen, we likely wouldn't be able to use native compilation there.

sqrl-planner/sqrl-planner-local/src/main/java/com/datasqrl/plan/global/DAGAssembler.java
henneberger: I'm just following the style that was already there. There is a lot going on in that method so I'll save it for a later refactor.

PR #272, opened 2023-05-30
Add graphql subscriptions

henneberger:
This adds subscriptions only accessible though tests. This feature is not yet supported for production level workloads but has a lot of benefit for automating testing. It is behind an inaccessible feature flag that is only accessible by tests.

Partially contributes to #273


mbroecheler (approved):
Just for the record: We should make it easier on the user and not require an explicit EXPORT statement for subscription but instead link the subscription to the table by name like we do with types.

PR #312, opened 2023-07-20
Fix idleness and file monitor issue

### LINKED ISSUE #308: Data not propagating after temporal joins for mixed sources
When query has a join between a log source and a file source, data may not propagate. Users may see data propagate for the first few seconds but then be unable to push more data.

For example:
IMPORT file.fileTable;
IMPORT schema.schemaTable;
Query := SELECT * FROM fileTable JOIN schemaTable;

A join operator will advance the smallest watermark seen on either side of the join. However, file data may never see new data after initial processing so the watermark for the query will never advance. Some strategies could be to have the file source emit a watermark after some moments of being idle, or emit a very large watermark when the file reaches the end. 

An idle timeout will be good for cases when the file monitor is used, and use large watermarks when processing a fixed set of files.

henneberger:
Fixes #308


sqrl-examples/banking/docker-compose.yml
```diff
@@ -25,25 +25,29 @@ services:
...
       - "8081:8081"
     command: jobmanager
+    user: root
     environment:
       - |
         FLINK_PROPERTIES=
         jobmanager.rpc.address: flink-jobmanager
+      - TZ:"America/Los_Angeles"
     volumes:
-      - /Users/matthias/git/sqml/sqrl-examples/banking:/Users/matthias/git/sqml/sqrl-examples/banking
```
mbroecheler, line 33: Is there a better way to solve this rather than hardcoding our respective file systems?

sqrl-execute/sqrl-execute-flink/sqrl-execute-flink-connectors/sqrl-execute-flink-file/src/main/java/com/datasqrl/io/FileSourceFactory.java
```diff
@@ -67,13 +72,20 @@ public SingleOutputStreamOperator<TimeAnnotatedRecord<String>> create(FlinkSourc
...
-        if (monitorDuration != null) {
-          builder.monitorContinuously(Duration.ofSeconds(10));
+
+        Optional<String> monitorInterval = tableConfig.getConnectorConfig()
+            .asString(FileSystemConnectorOptions.SOURCE_MONITOR_INTERVAL.key())
+            .getOptional();
+        if (isMonitor(monitorInterval)) {
+          Duration duration = monitorInterval
+              .map(TimeUtils::parseDuration)
+              .orElse(Duration.ofSeconds(10));
```
mbroecheler, line 87: Should we throw an exception if an invalid duration is configured?

sqrl-io/sqrl-io-file/src/main/java/com/datasqrl/io/impl/file/FileDataSystemConfig.java
```diff
@@ -27,6 +30,10 @@ public class FileDataSystemConfig implements Serializable {
   @Default
   String filenamePattern = DEFAULT_FILENAME_PATTERN;
 
+  @Default
+  @JsonProperty("source.monitor-interval")
+  String source$$monitor$interval = "10s";
```
mbroecheler, line 35: Why did you make this a nested property?
We should probably come up with a naming convention for properties so we don't confuse users with different styles (e.g. camelCase, dashes, underscores)

sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/cmd/DiscoverCommand.java
```diff
@@ -53,6 +54,7 @@ protected void runCommand(ErrorCollector errors) throws IOException {
     TableConfig discoveryConfig = null;
     if (inputFile != null && Files.isRegularFile(inputFile)) {
       discoveryConfig = TableConfig.load(inputFile, Name.system(inputFile.getFileName().toString()), errors);
+      applyConnectorOverrides(discoveryConfig.getConnectorConfig());
```
mbroecheler, line 57: Nice catch!

mbroecheler (commented):
See inline comments

sqrl-io/sqrl-io-file/src/main/java/com/datasqrl/io/impl/file/FileDataSystemConfig.java
henneberger: This is flink's table api config option for files. Some configuration options should be passthrough options, especially for the table api. This is particular example may be too much of a stretch since it's not passing through, but generally speaking we should support the full range of native configuration options.

PR #314, opened 2023-07-22
Update relnode plan conventions to support enumerable codegen

henneberger:


mbroecheler (approved):
Looks good if tests are stable. Created separate ticket to looks into why the index selection is broken (#319).

henneberger:
calcite planner replaces this

PR #318, opened 2023-07-28
[WIP] Separate flink function registration from table environment

henneberger:
First part in a series to remove flink hacks and isolate the flink table planner from the sqrl table planner.


henneberger:
calcite planner replaces this

Issue #334, opened 2023-08-29
Docker Compose Interpolation Format Error in Metrics Example


henneberger:
Encountering an error regarding the interpolation format for services.flink-job-submitter.command in the docker-compose.yml file. The exact error message was:

invalid interpolation format for services.flink-job-submitter.command: ""while ..... echo 'Job submitted.'"\n". You may need to escape any $ with another $

After following the suggestion and escaping $ with another $ in the docker-compose.yml file, it worked. 

Give a thumbs up if you are also facing this issue.

Environment:
OS: Apple M1 version 11.4
Docker version: 20.10.10, build b485636
Bash version: GNU bash, version 3.2.57(1)-release (arm64-apple-darwin20)

henneberger:
Solved in above commit

PR #335, opened 2023-08-29
Use alternative shell script for submitting flink

henneberger:
May resolve issue: #334


henneberger:
cherry picked into main. Released from tag

PR #367, opened 2023-10-11
Use ungeneralized types for flexible schema and planner

henneberger:
Moves away from the generalized graphql-esque type mapping to the more concrete sql types. 

Some schemas are published in the registry so the flexible schema is made to be backwards compatible.

Also removes UUID type.

### DIFF
--- a/sqrl-io/sqrl-io-schema-flexible/src/main/java/com/datasqrl/schema/type/basic/BasicType.java
+++ b/sqrl-io/sqrl-io-schema-flexible/src/main/java/com/datasqrl/schema/type/basic/BasicType.java
@@ -7,11 +7,8 @@ import com.datasqrl.schema.type.Type;
 import java.util.List;
 
 public interface BasicType<JavaType> extends Type, Comparable<BasicType<?>> {
-
   //First entry is the preferred name, the remaining are for backwards compatibility
-  List<String> getName();
-
+  String getName();
+  List<String> getNames();
   TypeConversion<JavaType> conversion();
-
-
 }
--- a/sqrl-planner/sqrl-common/src/main/java/com/datasqrl/calcite/QueryPlanner.java
+++ b/sqrl-planner/sqrl-common/src/main/java/com/datasqrl/calcite/QueryPlanner.java
@@ -194,25 +194,18 @@ public class QueryPlanner {
    * type definition syntax.
    */
   @SneakyThrows
-  public static RelDataType parseDatatype(String datatype) {
-    TypeFactory typeFactory = TypeFactory.getTypeFactory();
-    SqlCall sqlNode = (SqlCall) SqlParser.create(
-        String.format("CAST(null AS %s)", datatype)).parseExpression();
+  public RelDataType parseDatatype(String datatype) {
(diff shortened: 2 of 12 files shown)


sqrl-io/sqrl-io-schema-flexible/src/main/java/com/datasqrl/schema/type/basic/BasicType.java
```diff
@@ -4,10 +4,12 @@
...
 
 import com.datasqrl.schema.type.Type;
+import java.util.List;
 
-public interface BasicType<JavaType> extends Type, Comparable<BasicType> {
+public interface BasicType<JavaType> extends Type, Comparable<BasicType<?>> {
 
-  String getName();
+  //First entry is the preferred name, the remaining are for backwards compatibility
+  List<String> getName();
```
mbroecheler, line 12: IMHO a better approach would be to have two methods:
```
//Name of the type
default String getName() { getNames().get(0); }

//List of names for this type (for backwards compatibility)
List<String> getNames()
```

Otherwise you are leaking interface specifics (i.e. that the first name in the list is the default name) to the outside and pushing the responsibility onto the user to enforce it (as you can see by the various calls in AbstractBasicType to getName().get(0))

Also, a method that returns a collection should have a plural name.

sqrl-planner/sqrl-common/src/main/java/com/datasqrl/calcite/QueryPlanner.java
```diff
@@ -185,6 +189,30 @@ protected RelRoot planRoot(SqlValidator validator, SqlNode sqlNode) {
     return root;
   }
 
+  /**
+   * Helper function to parse a data type from a string using calcite's full
+   * type definition syntax.
+   */
+  @SneakyThrows
+  public static RelDataType parseDatatype(String datatype) {
```
mbroecheler, line 197: Should this go into CalciteUtil?

mbroecheler (changes requested):
Looks great, just one minor change please

sqrl-io/sqrl-io-schema-flexible/src/main/java/com/datasqrl/schema/type/basic/BasicType.java
henneberger: Good point, got too hung up on the generic issues and forgot about interface

henneberger: I'll actually tie it closer to the planner since it can be pretty expensive to run and the planner can hold the state we would need

sqrl-planner/sqrl-common/src/main/java/com/datasqrl/calcite/QueryPlanner.java
henneberger: I'll actually tie it closer to the planner since it can be pretty expensive to run and the planner can hold the state we would need

mbroecheler (approved):

PR #369, opened 2023-10-11
Add map support, fix array type information

henneberger:


sqrl-execute/sqrl-execute-flink/sqrl-execute-flink-connectors/sqrl-execute-flink-jdbc/src/main/java/org/apache/flink/connector/jdbc/internal/converter/BaseJdbcRowConverter.java
```diff
@@ -53,7 +54,9 @@ protected JdbcSerializationConverter wrapIntoNullableExternalConverter(
         }
       };
     } else if (type.getTypeRoot() == ROW) {
-      return (val, index, statement) -> {};
+      return (val, index, statement) -> setRow(type, val, index, statement);
+    }else if (type.getTypeRoot() == MAP) {
```
mbroecheler, line 58: whitespace

mbroecheler (changes requested):
Do we have a test case for maps and arrays being processed by flink and written to the database? How are those mapped in GraphQL?

Trying to understand what our story is for those complex types in this release (and beyond).

henneberger:
> Do we have a test case for maps and arrays being processed by flink and written to the database? How are those mapped in GraphQL?
> 
> Trying to understand what our story is for those complex types in this release (and beyond).

This makes it so we treat maps like other structured types, they get written as empty arrays and aren't surfaced in the graphql api. There is no primitive map type in postgres so it should be written in jsonb or other format.

A more comprehensive test suite is coming soon.

PR #391, opened 2023-10-28
Support table functions in SQL

henneberger:
Initial support for table functions in SQL. Supports calling sqrl table functions but does not impart sqrl native table paths. 

```
IMPORT ecommerce-data.Orders;
X(@id: Int) := SELECT * FROM Orders WHERE id = @id;
X(@id: Int, @customerid: Int) := SELECT * FROM Orders WHERE id = @id AND customerid = @customerid;
Y(@id: Int) := SELECT * FROM TABLE(X(2));
Z(@id: Int) := SELECT * FROM TABLE(X(2, 3));
```

```
IMPORT ecommerce-data.Orders;
X(@id: Int) := SELECT * FROM Orders WHERE id = @id;
Y(@id: Int) := SELECT * FROM TABLE(X(@id));
Z := SELECT * FROM TABLE(Y(3));
```

### DIFF
--- a/sqrl-planner/sqrl-common/src/main/java/org/apache/calcite/jdbc/SqrlSchema.java
+++ b/sqrl-planner/sqrl-common/src/main/java/org/apache/calcite/jdbc/SqrlSchema.java
@@ -85,7 +85,9 @@ public class SqrlSchema extends SimpleCalciteSchema {
     removePrefix(this.pathToAbsolutePathMap.keySet(),root.getName().toNamePath());
     plus().add(String.join(".", root.getPath().toStringList()) + "$"
         + sqrlFramework.getUniqueMacroInt().incrementAndGet(), root);
-    plus().add(root.getName().getDisplay(), root);
+    if (!root.getParameters().isEmpty()) {
+      plus().add(root.getName().getDisplay(), root);
+    }
   }
 
   private void removePrefix(Set<NamePath> set, NamePath prefix) {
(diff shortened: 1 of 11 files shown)


sqrl-planner/sqrl-common/src/main/java/org/apache/calcite/jdbc/SqrlSchema.java
```diff
@@ -107,6 +109,9 @@ public void addRelationship(Relationship relationship) {
     this.sysTableToRelationshipMap.put(relationship.getFromTable(), relationship);
     plus().add(String.join(".", relationship.getPath().toStringList()) + "$"
         + sqrlFramework.getUniqueMacroInt().incrementAndGet(), relationship);
+    if (!relationship.getParameters().isEmpty()) {
+      plus().add(String.join(".", relationship.getPath().toStringList()), relationship);
```
mbroecheler, line 113: This should call a helper function. We should already have one for this.

mbroecheler (changes requested):
This is awesome - thanks for getting this to work. It really highlights our contributions to SQL by being able to reuse those functions. Well done!
One minor nit.

sqrl-planner/sqrl-common/src/main/java/org/apache/calcite/jdbc/SqrlSchema.java
henneberger: Removed in next commit

PR #402, opened 2023-11-06
Generate SQL queries

mbroecheler:
…en't great.


henneberger:
Resigning this PR in favor of the approach of generating views for db engines by default.

PR #408, opened 2023-11-18
Add JSON library

henneberger:
1.  Adds JSON library
2. Removes some unnecessary type conversion logic, favoring flink's libraries
3. Modularizes function libraries
4. Conforms Vector library to new modular construction
5. Removes unnecessary type mapping between flink and sqrl, Sqrl now handles flink types natively
6. Fixes casting logic for database engines
7. We don't have a mapping between flink functions and their names that exist cross modularity boundaries. We can add dependencies but a more conservative approach was taken.

Notes about dependencies:
- Vertx w/ graphql-java v20 supports JsonObject as instrumentation, but right now we will need to use a vertx-dependent JSON type.
- Registering extensions should be service loaded in, which will let us drop vector as a dependency
- Haven't brought in flink as a dependency for vertx yet, so it's difficult to bridge flink types with graphql types. Same with execution libraries and engine/dialect mapping. We currently use Strings in some places where we should stricter types. Suggestions welcome.


sqrl-engines/sqrl-engine-flink/src/main/java/com/datasqrl/function/FlinkStdLibraryImpl.java
```diff
@@ -21,7 +21,22 @@ public class FlinkStdLibraryImpl extends AbstractFunctionModule implements StdLi
...
+      BuiltInFunctionDefinitions.JSON_EXISTS.getName(),
+      BuiltInFunctionDefinitions.JSON_OBJECT.getName(),
+      BuiltInFunctionDefinitions.JSON_QUERY.getName(),
+      BuiltInFunctionDefinitions.JSON_ARRAYAGG_ABSENT_ON_NULL.getName(),
+      BuiltInFunctionDefinitions.JSON_ARRAYAGG_NULL_ON_NULL.getName(),
+      BuiltInFunctionDefinitions.JSON_OBJECTAGG_ABSENT_ON_NULL.getName(),
+      BuiltInFunctionDefinitions.JSON_OBJECTAGG_NULL_ON_NULL.getName(),
+      BuiltInFunctionDefinitions.JSON_STRING.getName(),
+      BuiltInFunctionDefinitions.IS_JSON.getName(),
+      //Cast not yet supported, SQRL uses a String type that is not yet compatible
```
mbroecheler, line 38: Is there a ticket for this?

sqrl-engines/sqrl-engine-jdbc/src/main/java/com/datasqrl/engine/database/relational/JDBCEngine.java
```diff
@@ -161,16 +162,18 @@ private List<SqlDDLStatement> extractTypeExtensions(RelNode relNode) {
     Set<SqlDDLStatement> statements = new HashSet<>();
     //look at relnodes to see if we use a vector type
     for (RelDataTypeField field : relNode.getRowType().getFieldList()) {
-      if (field.getType() instanceof VectorType) {
+      if (field.getType() instanceof RawRelDataType &&
+        ((RawRelDataType) field.getType()).getRawType().getOriginatingClass() == FlinkVectorType.class)
```
mbroecheler, line 166: Can we please generalize this into the dialect? I.e. check in the dialect whether a particular type needs an extension?

sqrl-engines/sqrl-engine-jdbc/src/main/java/com/datasqrl/engine/database/relational/JDBCEngine.java
```diff
@@ -161,16 +162,18 @@ private List<SqlDDLStatement> extractTypeExtensions(RelNode relNode) {
...
         statements.add(new PostgresCreateVectorExtensionStatement());
-      }
     }
 
+    Set<String> vecFncs = Set.of("cosinesimilarity", "cosinedistance", "euclideandistance",
```
mbroecheler, line 170: Doing that by function name is very brittle imho. Instead, check the operand types generically. You can generalize this with the code above by pulling out all "non-standard" types first into a set and then add any extensions that a dialect might require.

sqrl-execute/sqrl-execute-http/sqrl-execute-http-core/src/main/java/com/datasqrl/graphql/server/BuildGraphQLEngine.java
```diff
@@ -31,27 +30,37 @@
...
 
-@NoArgsConstructor
 public class BuildGraphQLEngine implements
     RootVisitor<GraphQL, Context>,
     CoordVisitor<DataFetcher<?>, Context>,
     SchemaVisitor<TypeDefinitionRegistry, Object>,
     QueryBaseVisitor<ResolvedQuery, Context>,
     ResolvedQueryVisitor<CompletableFuture, QueryExecutionContext> {
 
+  private List<GraphQLScalarType> addlTypes;
```
mbroecheler, line 54: "additional" is not such a long word :-)

sqrl-execute/sqrl-execute-http/sqrl-execute-http-vertx/src/main/java/com/datasqrl/graphql/type/SqrlVertxScalars.java
```diff
@@ -0,0 +1,41 @@
...
+import graphql.schema.Coercing;
+import graphql.schema.GraphQLScalarType;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import lombok.SneakyThrows;
+
+public class SqrlVertxScalars {
+
+  static final ObjectMapper objectMapper = new ObjectMapper();
+
```
mbroecheler, line 14: This is where we would add a scalar type for vectors as well, right? Do we have a ticket for that somewhere?

sqrl-planner/sqrl-common/src/main/java/com/datasqrl/calcite/SqrlConfigurations.java
```diff
@@ -17,7 +17,7 @@ public class SqrlConfigurations {
       .withUpdateSetListNewline(false)
       .withIndentation(1)
       .withQuoteAllIdentifiers(true)
-      .withDialect(PostgresqlSqlDialect.DEFAULT)
+      .withDialect(ExtendedPostgresSqlDialect.DEFAULT)
```
mbroecheler, line 20: We need to be mindful to keep track of all the places where we hardcoded the dialect so we can generalize this soon. Nothing to change here right now. Just a general comment.

mbroecheler (changes requested):
Some minor change suggestions. The architecture looks much improved - great work on that. It's hard to review those changes in isolation, but the direction seems solid.
Caveat: I did not review the snapshots and assume that you reviewed those.

sqrl-engines/sqrl-engine-flink/src/main/java/com/datasqrl/function/FlinkStdLibraryImpl.java
henneberger: yeah: https://github.com/DataSQRL/sqrl/issues/409

sqrl-engines/sqrl-engine-jdbc/src/main/java/com/datasqrl/engine/database/relational/JDBCEngine.java
henneberger: Yeah, this was mentioned in PR message. `Registering extensions should be service loaded in`. I can do it now and drop the vector dependency, or make a new ticket.

sqrl-execute/sqrl-execute-http/sqrl-execute-http-vertx/src/main/java/com/datasqrl/graphql/type/SqrlVertxScalars.java
henneberger: Yeah, but I have it so it skips the field during graphql generation if it can't find the associated type (with a log line). A use can manually create the graphql field with a double array type and it would work.

PR #418, opened 2023-11-29
Premute only primary keys (#417)

### LINKED ISSUE #417: Permute primary key columns for initial graphql schema generation
Currently we add the first 8 fields of the schema as arguments to the fields of the base Query type, as well as the nested relationships fields. Since these fields are optional, we will permute all of the possible options. However, since the introduction of parameterized queries, this has become unnecessary since the user is able to define their query access patterns directly in SQRL. Instead, we will only add the primary key columns as columns in the generated schema. The permutation feature will still exist if no matching table function is found.

```graphql
type Query {
  Users(id: Float, first_name: String, last_name: String, email: String, ip_address: String, country: String, changed_on: Float, timestamp: String, limit: Int = 10, offset: Int = 0): [Users!]
}
 ```

becomes 
```graphql
type Query {
  Users(id: Float, limit: Int = 10, offset: Int = 0): [Users!]
}
 ```

### PR DESCRIPTION
**Pull request by henneberger**
Resolves #417

### DIFF
--- a/sqrl-planner/sqrl-planner-local/src/main/java/com/datasqrl/graphql/inference/SqrlSchemaForInference.java
+++ b/sqrl-planner/sqrl-planner-local/src/main/java/com/datasqrl/graphql/inference/SqrlSchemaForInference.java
@@ -5,8 +5,9 @@ import com.datasqrl.calcite.function.SqrlTableMacro;
 import com.datasqrl.canonicalizer.Name;
 import com.datasqrl.canonicalizer.NamePath;
 import com.datasqrl.canonicalizer.ReservedName;
+import com.datasqrl.plan.table.LogicalNestedTable;
+import com.datasqrl.plan.table.ScriptRelationalTable;
 import com.datasqrl.schema.Multiplicity;
-import com.datasqrl.schema.Relationship;
 import com.datasqrl.schema.Relationship.JoinType;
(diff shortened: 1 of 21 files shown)

### CODE REVIEW

File: sqrl-planner/sqrl-planner-local/src/main/java/com/datasqrl/graphql/inference/SqrlSchemaForInference.java [hunk 1]
```diff
@@ -72,18 +72,38 @@ private SQRLTable createSqrlTable(SqrlTableMacro macro, SqrlSchema schema) {
...
+          field.getName(),
+          isLocalPrimaryKey(table, i)));
     }
     return sqrlTable;
   }
 
+  private boolean isLocalPrimaryKey(Table table, int i) {
+    if (table instanceof LogicalNestedTable) {
+      LogicalNestedTable nestedTable = (LogicalNestedTable) table;
+      if (i < nestedTable.getNumPrimaryKeys() - nestedTable.getNumLocalPks()) {
```
mbroecheler, line 92: `return i >= nestedTable.getNumPrimaryKeys() - locakPks() && i< getNumPrimaryKeys()`

mbroecheler (approved):
The GraphQL schemas look so much better. This is a much needed simplification. Minor nit.

PR #427, opened 2023-12-06
Copy S3 plugin to flink for docker compose

henneberger:
Follows the recommended practice as outlined here:

https://nightlies.apache.org/flink/flink-docs-stable/docs/deployment/filesystems/plugins/

This copies the presto s3 connector. If using azure, other plugins will need to be copied in. To use, set the aws access key and secret key like the following:

```
  flink-jobmanager:
    image: flink:1.16.1-scala_2.12-java11
    ports:
      - "8081:8081"
    command: /bin/bash /exec/init-flink.sh jobmanager
    environment:
      - |
        FLINK_PROPERTIES=
        jobmanager.rpc.address: flink-jobmanager
      - AWS_ACCESS_KEY_ID=...
      - AWS_SECRET_KEY=...
    volumes:
      - ./init-flink.sh:/exec/init-flink.sh
```

### DIFF
--- a/sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/compile/DockerCompose.java
+++ b/sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/compile/DockerCompose.java
@@ -7,14 +7,13 @@ import java.util.Optional;
 public class DockerCompose {
 
   public static String getYml(Optional<Path> mountDir) {
-    String volumneMnt = mountDir.map(dir -> dir.toAbsolutePath().normalize())
-        .map(dir -> ""
-            + "    volumes:\n"
-            + "      - " + dir.toAbsolutePath() + ":/build\n"
-            + "      - ./init-flink.sh:/exec/init-flink.sh\n")
-        .orElse(""
-            + "    volumes:\n"
-            + "      - ./init-flink.sh:/exec/init-flink.sh\n");
+    String volumneMnt = ""
+        + "    volumes:\n"
+        + mountDir
+        .map(dir -> dir.toAbsolutePath().normalize())
+        .map(dir -> "      - " + dir.toAbsolutePath() + ":/build\n")
+        .orElse("")
+        + "      - ./init-flink.sh:/exec/init-flink.sh\n";
 
     return "# This is a docker-compose template for starting a DataSQRL compiled data pipeline\n"
         + "# This template uses the Apache Flink as the stream engine, Postgres as the database engine, and Vertx as the server engine.\n"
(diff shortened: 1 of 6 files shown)


sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/compile/DockerCompose.java
```diff
@@ -8,8 +8,13 @@ public class DockerCompose {
...
     String volumneMnt = mountDir.map(dir -> dir.toAbsolutePath().normalize())
-        .map( dir -> "    volumes:\n"
-        + "      - " + dir.toAbsolutePath() + ":/build\n").orElse("");
+        .map(dir -> ""
+            + "    volumes:\n"
+            + "      - " + dir.toAbsolutePath() + ":/build\n"
+            + "      - ./init-flink.sh:/exec/init-flink.sh\n")
+        .orElse(""
+            + "    volumes:\n"
+            + "      - ./init-flink.sh:/exec/init-flink.sh\n");
```
mbroecheler, line 17: Can you make that a static string for the init-flink and then append the optional dir so you don't have to repeat the string?

sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/compile/DockerCompose.java
```diff
@@ -104,6 +109,15 @@ public static String getYml(Optional<Path> mountDir) {
         + "    driver: local";
   }
 
+  public static String getInitFlink() {
+    return "#!/bin/bash\n"
+        + "\n"
+        + "# Copy the JAR file to the plugins directory\n"
+        + "mkdir -p /opt/flink/plugins/s3-fs-presto\n"
+        + "cp /opt/flink/opt/flink-s3-fs-presto-1.16.1.jar /opt/flink/plugins/s3-fs-presto/\n"
```
mbroecheler, line 117: so, this is where we would have to copy all the plugins we end up using?

mbroecheler (approved):
One minor nit

sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/compile/DockerCompose.java
henneberger: yeah

sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/compile/DockerCompose.java
henneberger: fixed so it's nicer

PR #430, opened 2023-12-07
Use char(36) for UUID function output

henneberger:


mbroecheler (approved):
Uf, this is a tough one to review. I'm not sure about this change for all the butterfly effects it could cause? Is it going to make SQRL sentient and destroy humanity???

PR #461, opened 2023-12-14
Add readme to udf example, fix minor udf related bugs

henneberger:


sqrl-examples/udf/README.md
```diff
@@ -0,0 +1,62 @@
...
+4. **Jar Compiling:** Compile the sample project and build a jar. This jar is what DataSQRL will use to discover your function. It reads the manifest entries for any ScalarFunction or AggregateFunction classes and load them into DataSQRL for use in queries. It can be placed into any folder relative to the sqrl root folder which will translate to the import path. In the example, we will use the `target` directory that the compilation process creates.
+
+## SQRL Compilation and Packaging
+1. **SQRL Compilation:** Compile the SQRL using DataSQRL's command interface, which prepares your script for deployment in the Flink environment.
+
+```shell
+docker run --rm -v $PWD:/build datasqrl/cmd compile myudf.sqrl schema.graphqls --mnt $PWD
+```
+
+2. **Alternate Packaging:** Next, you need to package the jar using a packager that will repack your library with the deployable flink jar. If you ship your jar to flink directly, you can skip this step. This step can take some time to compile.
```
mbroecheler, line 25: I would reword this as there being two options for deploying the jar:
1. Package it with the Flink jar into one deployable jar. Benefit: It's one jar that has all dependencies and can be deployed to any Flink cluster. Downside: extra step to package one combined jar
2. Upload it to the Flink cluster directly. Benefit: every Flink job running on that cluster can now use the function. Downside: you have to upload the cluster which requires access.

sqrl-examples/udf/schema.graphqls
```diff
@@ -0,0 +1,20 @@
+type Query {
```
mbroecheler, line 1: I like the naming convention of giving the sqrl script and the schema the same name. Makes it easier to see what goes together. (i.e. myudf.graphqls)

mbroecheler (approved):
Some minor suggestions

sqrl-examples/udf/README.md
henneberger: I'm actually going to remove the part about uploading the jar directly since it doesn't seem to be mentioned in the flink documentation so it may be non-standard. The docs all talk about bundling into a fat jar.

PR #472, opened 2024-01-08
changed LocalDateTime to ZonedDateTime

hubchub:
The test previously only worked if you were in the PST - America/Los_Angeles time zone but it now passes regardless of where the script is run.

### DIFF
--- a/sqrl-execute/sqrl-execute-flink/sqrl-execute-flink-function/src/main/java/com/datasqrl/TimeFunctions.java
+++ b/sqrl-execute/sqrl-execute-flink/sqrl-execute-flink-function/src/main/java/com/datasqrl/TimeFunctions.java
@@ -372,15 +372,13 @@ public class TimeFunctions {
     }
 
     public Instant eval(String s, String format) {
-      DateTimeFormatter formatter = DateTimeFormatter.ofPattern(format, Locale.US);
-      try {
-        return LocalDateTime.parse(s, formatter)
-            .atZone(ZoneId.systemDefault())
-            .toInstant();
-      } catch (Exception e) {
-        log.warn(e.getMessage());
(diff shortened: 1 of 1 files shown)


sqrl-execute/sqrl-execute-flink/sqrl-execute-flink-function/src/main/java/com/datasqrl/TimeFunctions.java
```diff
@@ -372,15 +372,13 @@ public Instant eval(String s) {
...
-        return LocalDateTime.parse(s, formatter)
-            .atZone(ZoneId.systemDefault())
-            .toInstant();
-      } catch (Exception e) {
-        log.warn(e.getMessage());
-        return null;
-      }
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern(format, Locale.US);
+        try {
+            return ZonedDateTime.parse(s, formatter).toInstant();
```
henneberger, line 377: Thanks for the contribution! We're all PST guys so this feedback has been very important to us.

Earlier we made the decision to normalize our dates for stream processing engines. This was so time-based operations like windowing, joining, or aggregating could remain consistent. The ZonedDateTime approach (which is superior and we should move to), would require a `.withZoneSameInstant(ZoneId.systemDefault())` for the same normalization, but we end up in locality quirks are you were observing. I'm not 100% sure if our assumption about date normalization is still valid and watermarking quirks can be very insidious, but it's been awhile since I've looked at it. Perhaps @mbroecheler has some input, or I can take a closer look at it later today.

henneberger:
Closing as the branch has gone stale. We will still use local time as the default time zone, but we may want to consider using UTC

PR #473, opened 2024-01-09
Credit Card Example

mbroecheler:


sqrl-examples/src/main/java/com/datasqrl/data/GenerateCreditCards.java
```diff
@@ -0,0 +1,254 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import lombok.Value;
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "creditcard", description = "Generates Credit Card data")
+public class GenerateCreditCards extends AbstractGenerateCommand {
```
henneberger, line 23: How do you feel about adding these to the creditcard example folder instead of having sqrl-examples have a class root? Examples should be reasonably self contained so they are easily reproducible for users.

PR #493, opened 2024-01-29
install pgvector extension for postgres

wutianchen:
version 14, verified inside container with psql

Screenshot 2024-01-29 at 14 18 17

### DIFF
--- a/sqrl-tools/sqrl-cli/Dockerfile
+++ b/sqrl-tools/sqrl-cli/Dockerfile
@@ -1,15 +1,24 @@
 FROM eclipse-temurin:11
 WORKDIR /usr/src/app
+ARG POSTGRES_VERSION=14
 
 # Install PostgreSQL
-RUN apt-get update && apt-get install -y postgresql postgresql-contrib && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y postgresql postgresql-contrib lsb-release gnupg && rm -rf /var/lib/apt/lists/*
 
+# Install pgvector extension package
+RUN sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'  && \
+    wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -  && \
+    apt-get update && \
+    apt-get install -y postgresql-${POSTGRES_VERSION}-pgvector && \
+    rm -rf /var/lib/apt/lists/*
+
+#postgresql-16-pgvector
 RUN service postgresql start && \
     su - postgres -c "psql -U postgres -c \"ALTER USER postgres WITH PASSWORD 'postgres';\"" && \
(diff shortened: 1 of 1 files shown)


sqrl-tools/sqrl-cli/Dockerfile
```diff
@@ -1,15 +1,25 @@
 FROM eclipse-temurin:11
 WORKDIR /usr/src/app
+ARG POSTGRES_VERSION=14
```
henneberger, line 3: I see version 14 here but version 16 down below. Is this intended?

henneberger: Also, the arg isn't required, we should just fix to the version that the container installs

sqrl-tools/sqrl-cli/Dockerfile
```diff
@@ -1,15 +1,25 @@
 FROM eclipse-temurin:11
```
henneberger, line 1: Let's fix this docker image to a more specific version so the postgres dependency remains consistent over time

sqrl-tools/sqrl-cli/Dockerfile
wutianchen: thanks for the message. I fixed the comment. I installed the same version as postgres in docker container. 

As for the ARG, I try to parameterize the version. later on also use that arg probably for postgres installation. How do you think about that ?

wutianchen: so that pgvector version always match with postgresql installation

sqrl-tools/sqrl-cli/Dockerfile
wutianchen: okay. you mean to fix the version of eclipse-temurin base image ? which specific version do you suggests ? thanks

henneberger: I think it's [11.0.22_7-jdk](https://hub.docker.com/layers/library/eclipse-temurin/11.0.22_7-jdk/images/sha256-8aa05793d70544e950e50f35f3c97c8a7ba369bf132fb28364bb9d34e1129478?context=explore)

sqrl-tools/sqrl-cli/Dockerfile
henneberger: Instead of downgrading to 14, can we use postgres 15 or higher? Our test coverage is around pg 15.

henneberger: Yeah, it'd be nice to have the pgvector version match the pg version.

wutianchen: ok, will try to upgrade to v15

sqrl-tools/sqrl-cli/Dockerfile
wutianchen: done

sqrl-tools/sqrl-cli/Dockerfile
wutianchen: done

henneberger (approved):

PR #503, opened 2024-02-01
Update json functions to handle nulls

### LINKED ISSUE #502: Some json functions fail on null values
Using json functions in the standard json library will fail with a null pointer exception.

henneberger:
Fixes #502


mbroecheler (approved):
ChatGPT is great at generating test cases ;-)

PR #508, opened 2024-02-08
0.5 Refactor

mbroecheler:
- expose more source configuration in table config
- simplify LPRewriter
- improve join handling
- simplify timestamp handling (and make it deterministic)


sqrl-planner/sqrl-planner-local/src/main/java/com/datasqrl/plan/table/QueryRelationalTable.java
```diff
@@ -19,12 +20,12 @@ public class QueryRelationalTable extends PhysicalRelationalTable {
...
-  public QueryRelationalTable(Name rootTableId, Name tableName, @NonNull LPAnalysis analyzedLP) {
-    super(rootTableId, tableName,
+  public QueryRelationalTable(Name rootTableId, NamePath tablePath, @NonNull LPAnalysis analyzedLP) {
+    super(rootTableId, tablePath,
         analyzedLP.getConvertedRelnode().getType(),
         analyzedLP.getConvertedRelnode().getRelNode().getRowType(),
-        analyzedLP.getConvertedRelnode().getTimestamp().finalizeAsBase(),
-        analyzedLP.getConvertedRelnode().getPrimaryKey().getLength(),
+        analyzedLP.getConvertedRelnode().getTimestamp(),
+        PrimaryKey.of(analyzedLP.getConvertedRelnode().getPrimaryKey()),
```
henneberger, line 28: It'd probably be nicer to back this with a RelOptTable rather than a calcite Table, but we can do that down the road

PR #510, opened 2024-02-10
Add optional engines

henneberger:


sqrl-planner/sqrl-planner-local/src/main/java/com/datasqrl/plan/global/DAGBuilder.java
```diff
@@ -51,24 +51,30 @@ public SqrlDAG build(Collection<AnalyzedAPIQuery> queries,
     Multimap<SqrlNode, SqrlNode> dagInputs = HashMultimap.create();
     //1. Add all queries as sinks
     List<ExecutionStage> readStages = pipeline.getReadStages();
-    errors.checkFatal(!readStages.isEmpty(), "Configured Pipeline does not include"
-        + " any read stages to execute API queries: %s",pipeline);
+    if (readStages.isEmpty()) {
```
mbroecheler, line 54: Can you make this a !readStages.isEmpty() check around the for loop so we don't have to repeat that code?

sqrl-planner/sqrl-planner-local/src/main/java/com/datasqrl/plan/global/DAGBuilder.java
```diff
@@ -51,24 +51,30 @@ public SqrlDAG build(Collection<AnalyzedAPIQuery> queries,
...
+    export(exports, table2Node, dagInputs);
+    return new SqrlDAG(dagInputs);
+  }
+
+  private void export(Collection<ResolvedExport> exports,
+      Map<PhysicalRelationalTable, TableNode> table2Node, Multimap<SqrlNode, SqrlNode> dagInputs) {
     int numExports = 1;
     List<ExecutionStage> exportStages = pipeline.getStages().stream().filter(s -> s.supports(
         EngineCapability.EXPORT)).collect(Collectors.toList());
-    errors.checkFatal(!exportStages.isEmpty(), "Configured Pipeline does not include "
```
mbroecheler, line 64: Shouldn't we check if the exports are actually empty before we return? Meaning, what if there are exports and no stage that supports them?

mbroecheler (approved):
Looks great. Just some minor suggestions. Thanks for adding thorough test cases!

PR #515, opened 2024-02-13
Fix automatic downcast bug for connectors without native json support

henneberger:


sqrl-engines/sqrl-engine-flink/src/main/java/com/datasqrl/engine/stream/flink/plan/SqrlToFlinkExecutablePlan.java
```diff
@@ -195,30 +198,36 @@ private static RexNode convertField(RelDataTypeField field, AtomicBoolean hasCha
...
-            .getName());
+      if (downcastFunction == null) {
+        throw new RuntimeException("Needed downcast function but could not find one: " + defaultConversion.getName());
       }
-    }
+      String fncName = downcastFunction.downcastFunctionName().toLowerCase();
+
+      List<SqlOperator> operators = new ArrayList<>();
+      catalogReader.getSchema().getSqrlFramework()
+          .getSqrlOperatorTable().lookupOperatorOverloads(
```
henneberger, line 216: This is still incorrect, the downcast function may not have been imported. Probably a reflection construction is best here since we'll need the return type for proper inference.

PR #519, opened 2024-02-15
Clean up function dependencies

henneberger:


mbroecheler (approved):
Did not look at this too closely. I assume you broke up the function classes into individual classes per function and that this remains functionally equivalent. Please note that I redid the marker interfaces for functions a bit in my 0.5 branch - just a heads up as you rebase.

PR #523, opened 2024-02-17
Add native json handling to json format

henneberger:


sqrl-engines/sqrl-engine-flink/src/main/java/com/datasqrl/engine/stream/flink/plan/SqrlToFlinkExecutablePlan.java
```diff
@@ -197,51 +193,57 @@ private RelNode applyDowncasting(RelNode relNode, Optional<ExecutionEngine> engi
...
-    return relBuilder.field(field.getIndex());
+    downcastClassNames.put(fncName,
+        downcastFunction.getDowncastClassName().getName());
+
+    hasChanged.set(true);
+    return relBuilder.getRexBuilder()
+        .makeCall(convertedFunction.get(), List.of(relBuilder.field(field.getIndex())));
+  }
+
+  private boolean engineSupportsType(Optional<ExecutionEngine> engine, Class<?> defaultConversion) {
```
mbroecheler, line 245: Just some food for thought: I see that you are tying the determination of downcasting to the engine. From the perspective of Flink, I wonder if we should make that determination for the connector.
In other words, does it matter whether we are writing to Postgres as an engine or to Postgres as an external connector?

mbroecheler (approved):
Looks good, thank you

sqrl-engines/sqrl-engine-flink/src/main/java/com/datasqrl/engine/stream/flink/plan/SqrlToFlinkExecutablePlan.java
henneberger: yeah, we can move to a connector + format when we transition away from the connectors that are streams. Right now it's a lot of code to migrate.

PR #524, opened 2024-02-17
Allow any object type for objectagg

henneberger:


sqrl-lib/sqrl-json/sqrl-json-execute/src/main/java/com/datasqrl/json/JsonFunctions.java
```diff
@@ -474,6 +476,12 @@ public void accumulate(ObjectAgg accumulator, String key, Integer value) {
       accumulateObject(accumulator, key, value);
     }
 
+    @FunctionHint(input = {@DataTypeHint(bridgedTo = String.class),
+        @DataTypeHint(inputGroup = InputGroup.ANY)})
+    public void accumulate(ObjectAgg accumulator, String key, Object value) {
```
henneberger, line 481: I'll need to check how this serializes in other cases. It does work for the trivial example in flink

PR #525, opened 2024-02-18
Migrate to JsonNode for Json functions

henneberger:


mbroecheler (approved):
Didn't review it too carefully. Looks like it mostly substitutes the string wrapped by FlinkJsonType with JsonNode which makes sense.
I wonder if there is any downside to moving JsonNode around like that but that's for a later day ;-)

henneberger:
This cannot be merged in yet because JsonNode cannot be serialized when submitting in a jar.

henneberger:
This ended up misguided since we should specify the serializer on the type

PR #535, opened 2024-03-21
Allow arbitrary length json during serialization

henneberger:


henneberger:
Ah, this ended up in another commit

PR #538, opened 2024-03-27
Add integration test example

henneberger:
Testing the build to see if the integration tests in docker will run


henneberger:
We can bring this back when we develop the sqrl run command more

PR #541, opened 2024-04-05
Favor local profile during repository resolution

### LINKED ISSUE #540: Package bootstrap attempts repository resolution on paths
The package bootstrap (-p switch) allows either a remote repository identifier or a path identifier. We should check if a local profile exists first and favor that.

The SQRL documentation should also be updated.

https://github.com/DataSQRL/sqrl/blob/8008f4a9b7693c5e0eec094101a9b7dae23b90aa/sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/cmd/PackageBootstrap.java#L93

https://github.com/DataSQRL/sqrl/blob/1cabe80049568081f8228efad8787ba4fa27b0ce/sqrl-tools/sqrl-packager/src/main/java/com/datasqrl/packager/repository/CompositeRepositoryImpl.java#L27-L26

nandorsoma:
Fixes #540

### DIFF
--- a/sqrl-tools/sqrl-cli/pom.xml
+++ b/sqrl-tools/sqrl-cli/pom.xml
@@ -60,13 +60,11 @@
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
-      <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-junit-jupiter</artifactId>
-      <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/sqrl-tools/sqrl-cli/pom.xml
+++ b/sqrl-tools/sqrl-cli/pom.xml
@@ -58,6 +58,18 @@
     </dependency>
     <!-- Test jars -->
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>${mockito.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-junit-jupiter</artifactId>
+      <version>${mockito.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <artifactId>sqrl-calcite</artifactId>
       <groupId>run.sqrl</groupId>
       <type>test-jar</type>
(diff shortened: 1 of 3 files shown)


sqrl-tools/sqrl-cli/pom.xml
```diff
@@ -57,6 +57,18 @@
       <scope>runtime</scope>
     </dependency>
     <!-- Test jars -->
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>${mockito.version}</version>
```
henneberger, line 63: Remove the version field. The version field should live in the root pom and these dependency versions managed by the dependencyManagement section

sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/cmd/PackageBootstrap.java
```diff
@@ -192,6 +185,15 @@ public SqrlConfig bootstrap(Repository repository, ErrorCollector errors,
     return postProcess.apply(sqrlConfig);
   }
 
+  // Currently the compilation fails with a nice stack trace when there are multiple package.jsons
+  // Normally we want to merge these jsons.
+  // This method is a temporary solution to give a meaningful error message to the user until we fix this behavior.
+  private void detectDuplicateConfigs(ErrorCollector errors, List<Path> configFiles) {
+    if (configFiles.stream().anyMatch(path -> path.equals(rootDir.resolve(DEFAULT_PACKAGE)))) {
+      errors.fatal("Both the project and the profile have a package.json defined.");
```
henneberger, line 193: This check can be dropped for a mutable configFiles. Being able to override a profile's package json will be important for flexibility and we should attempt to preserve that functionality.

henneberger:
Look great 👍! Just a minor suggestion.

sqrl-tools/sqrl-cli/pom.xml
nandorsoma: Yeah, makes sense. Thanks!

sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/cmd/PackageBootstrap.java
nandorsoma: So, originally, I wanted to include this check to provide better clarity since this scenario is already failing. However, upon reconsideration, I realize it might actually create more issues than it resolves. Therefore, I've decided to remove it.

nandorsoma:
Thanks for the review @henneberger, please see my latest changes.

henneberger (approved):
lgtm!

PR #543, opened 2024-04-08
Make configFiles list mutable, fix testRetailPackaging

nandorsoma:


sqrl-tools/sqrl-packager/src/main/java/com/datasqrl/packager/Packager.java
```diff
@@ -294,7 +293,7 @@ public static Optional<List<Path>> findPackageFile(Path rootDir, List<Path> pack
     if (packageFiles.isEmpty()) {
       Path defaultPkg = rootDir.resolve(DEFAULT_PACKAGE);
       if (Files.isRegularFile(defaultPkg)) {
-        return Optional.of(List.of(defaultPkg));
+        return Optional.of(Lists.newArrayList(defaultPkg));
```
henneberger, line 296: It's probably a safer assumption to assume that any list being passed between functions is immutable or assumed to be immutable. We should just create a mutable list and populate it at the point of mutation.

henneberger:
closing since we should not assume lists are mutable.

PR #559, opened 2024-04-30
Support for repository v2

nandorsoma:


sqrl-tools/sqrl-packager/src/main/java/com/datasqrl/packager/repository/RemoteRepositoryImplementation.java
```diff
@@ -30,14 +22,14 @@
...
+import lombok.Setter;
+import net.lingala.zip4j.ZipFile;
+import org.apache.commons.io.FileUtils;
 
 public class RemoteRepositoryImplementation implements Repository {
-  public static final String PKG_NAME_KEY = "pkgName";
-  public static final String VERSION_KEY = "version";
-  public static final String VARIANT_KEY = "variant";
-
   public static final URI DEFAULT_URI = URI.create("https://repo.sqrl.run");
```
nandorsoma, line 33: Currently we need to manually change it to the staging url to make this work. What should we do here? Are we going to forward this url to the new repo once we release the new version?

sqrl-tools/sqrl-auth/src/main/java/com/datasqrl/auth/AuthProvider.java
```diff
@@ -0,0 +1,155 @@
...
+    public AuthProvider(HttpClient client) {
+        this.client = client;
+    }
+
+    @SneakyThrows
+  public String loginToRepository() {
+    CompletableFuture<String> tokenFuture = new CompletableFuture<>();
+
+    Vertx vertx = Vertx.vertx();
+    vertx.deployVerticle(new OAuthCallbackVerticle(code -> exchangeCodeForToken(code, tokenFuture)));
```
henneberger, line 52: Vertx is heavyweight. I'd recommend using java's HttpServer, which is enough to handle a callback

sqrl-tools/sqrl-auth/src/main/java/com/datasqrl/auth/AuthProvider.java
```diff
@@ -0,0 +1,155 @@
...
+    openAuthWindow();
+
+    try {
+      String authToken = tokenFuture.get(TIMEOUT_IN_SECONDS, TimeUnit.SECONDS);
+
+      Files.createDirectories(DATASQRL_CONFIG_DIRECTORY);
+
+      Files.write(AUTH_TOKEN_PATH, authToken.getBytes(StandardCharsets.UTF_8));
+
+      return tokenFuture.get(TIMEOUT_IN_SECONDS, TimeUnit.SECONDS);
```
henneberger, line 63: 60 second to log in is a rush. Why not just hold until the window closes?

sqrl-tools/sqrl-auth/src/main/java/com/datasqrl/auth/AuthProvider.java
```diff
@@ -0,0 +1,155 @@
...
+            .collect(Collectors.joining("&"));
+
+    String authUrl = "https://" + AUTH0_DOMAIN + "/authorize?" + paramString;
+
+    Runtime.getRuntime().exec("open " + authUrl);
+  }
+
+  @SneakyThrows
+  private boolean isTokenValid(String authToken) {
+    HttpRequest request = HttpRequest.newBuilder()
```
henneberger, line 131: Do jwt token validation instead since it's suppose to be stateless.

sqrl-tools/sqrl-auth/src/main/java/com/datasqrl/auth/AuthProvider.java
```diff
@@ -0,0 +1,155 @@
...
+      return tokenFuture.get(TIMEOUT_IN_SECONDS, TimeUnit.SECONDS);
+    } catch (TimeoutException | ExecutionException | InterruptedException e) {
+      throw new RuntimeException("Exception while getting access token.", e);
+    } finally {
+      vertx.close();
+    }
+  }
+
+  @SneakyThrows
+  public String isAuthenticated() {
```
henneberger, line 72: This method signature is awkward to return a string for isAuth. What about something like Optional<JwtToken> getToken() or boolean isAuth()

sqrl-tools/sqrl-packager/src/main/java/com/datasqrl/packager/repository/RemoteRepositoryImplementation.java
```diff
@@ -115,68 +99,69 @@ private boolean downloadDependency(Path targetPath, JsonNode dep, Dependency dep
...
-    if (!packages.isArray() || packages.isEmpty()) return Optional.empty();
-    return Optional.of(packages.get(0).get(field));
-  }
-
-  public JsonNode executeQuery(Query query, Map<String,Object> payload) {
+  public JsonNode getDependencyInfo(String name, String version, String variant) {
     try {
       HttpClient client = HttpClient.newHttpClient();
+
+      String authToken = authProvider.isAuthenticated();
```
henneberger, line 110: Shouldn't we also have a refresh token workflow too?

sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/cmd/PublishCommand.java
```diff
@@ -20,32 +21,37 @@
 @CommandLine.Command(name = "publish", description = "Publishes a package to local and remote repository")
 public class PublishCommand extends AbstractCommand {
 
-    @CommandLine.Parameters(index = "0", description = "Main script (optional)")
+    @CommandLine.Option(names = {"--main"}, description = "Main script (optional)")
```
henneberger, line 24: new config switch? won't this break our docs?

sqrl-tools/sqrl-auth/src/main/java/com/datasqrl/auth/AuthProvider.java
henneberger, line 52: should be in the try block so vertx can clean up if something goes wrong with deploying the verticle or opening the browser

sqrl-tools/sqrl-auth/src/main/java/com/datasqrl/auth/AuthProvider.java
nandorsoma: Do you want to listen to an event when the user closes the browser window?

sqrl-tools/sqrl-auth/src/main/java/com/datasqrl/auth/AuthProvider.java
nandorsoma: As we agreed, we can revisit this topic in a followup, but for now it will stay as it is.

PR #593, opened 2024-05-08
Add support for graphql-ws websockets

### LINKED ISSUE #507: Upgrade websocket to graphql-ws
subscriptions-transport-ws is deprecated

henneberger:
Closes #507

By convention, graphql-ws uses the same route as the graphql endpoint, so it is no longer configurable.


mbroecheler (approved):
👍

PR #597, opened 2024-05-10
Validate package subsection in package.json

nandorsoma:
This PR introduces validation for config files, with initial focus on validating packages being uploaded to the repository. The ultimate goal is to make it extendable for the validation of the entire Sqrl config, hence its placement within _SqrlConfigCommons_. Here are some considerations driving the current design:

- **Individual vs. combined configurations:** We can validate either individual configurations or the resultant `CombinedConfiguration`. While the `CombinedConfiguration` is critical for us, providing validation feedback on individual files may be more beneficial to the user. Hence, every file is validated separately.
  
- **Precedence in configurations:** Given that if a section is present across multiple files, one of them takes precedence over the others, we need to consider if this is valid for the "package" section. Should there be only one "package" section? If so, we may require a "strict" combination rule which doesn't allow overrides for certain use-cases. This requires further thoughts and feedback.
  
- **Validation point:** Validation could be performed either before parsing the file into `Configuration` (validating raw user input) or after (validating `Configuration` object). The latter avoids duplicate file reads from the disk and has been implemented for now.
  
- **Configuration to JsonObject:** The conversion of `Configuration` to `JsonObject` required a custom converter. An alternate approach would utilize `SqrlConfigCommons.toMap` and pass the result to the `JsonObject` constructor, necessitating minor modifications to the `toMap` method. This alternate approach was not chosen to avoid touch-points with `getKeys` and `getAllKeys`, and to retain code readability despite the minor duplication in function.

I look forward to further discussions on these points, particularly on the potential implementation of a "strict" combination rule and the preferred approach for converting `Configuration` to `JsonObject`.

### DIFF
--- a/sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfigCommons.java
+++ b/sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfigCommons.java
@@ -18,10 +18,12 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -29,11 +31,16 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.function.Predicate;
+
+import com.google.common.io.Resources;
+import io.vertx.core.json.JsonObject;
+import io.vertx.json.schema.*;
 import lombok.AccessLevel;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 import lombok.NonNull;
+import lombok.SneakyThrows;
 import org.apache.commons.configuration2.BaseHierarchicalConfiguration;
 import org.apache.commons.configuration2.CombinedConfiguration;
 import org.apache.commons.configuration2.Configuration;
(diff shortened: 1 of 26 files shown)


sqrl-tools/sqrl-config/src/main/resources/schema/package-schema.json
```diff
@@ -0,0 +1,56 @@
...
+  "properties": {
+    "package": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "version": {
+          "type": "string",
+          "pattern": "^(\\d+\\.\\d+\\.\\d+)$"
```
nandorsoma, line 13: Do we want to enforce a pattern in versioning or is it enough if it is a string?

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfigCommons.java
```diff
@@ -344,28 +351,70 @@ public static SqrlConfig fromFiles(ErrorCollector errors, @NonNull Path firstFil
...
-      return
-          new SqrlConfigCommons(errors.withConfig(configFilename), configFilename, resultconfig, "");
+      return new SqrlConfigCommons(errors.withConfig(configFilename), configFilename, resultconfig, "");
     } catch (ConfigurationException e) {
       throw localErrors.handle(e);
     }
   }
 
+  @SneakyThrows
+  private static boolean isValid(Configuration baseconfig) {
```
mbroecheler, line 380: SqrlConfigCommons is a general configuration class that we use to read all kinds of configuration files. The logic for validating that a file has a particular schema needs to live in a separate class to not mix concerns.

sqrl-tools/sqrl-config/src/main/resources/schema/package-schema.json
```diff
@@ -0,0 +1,58 @@
...
+            "type": "string"
+          }
+        }
+      },
+      "required": [ "name", "version", "latest", "keywords" ],
+      "additionalProperties": false
+    }
+  },
+  "required": [ "package" ],
+  "additionalProperties": false
```
mbroecheler: The package.json file is used to configure the datasqrl compiler. It contains lots of other configuration options.
I'm surprised that this does not break the compiler. Are you not getting test failures with this?

mbroecheler (changes requested):
Did a quick cursory review. Not sure if this PR is still relevant but thought the comments might help to understand the context.

henneberger:
closing due to staleness. we can readdress later since it is still needed

Issue #581, opened 2024-05-08
Slim down CLI jar


henneberger:
There are unused dependencies in the flink jar that can be removed. We used to use kafka and flink more extensively for the 'run' command but that has been removed.

PR #610, opened 2024-05-15
Slim down CLI jar

nandorsoma:
I removed all dependencies that weren't being used in the project. Unfortunately, this didn't significantly reduce the size; it only reduced the jar size by 17MB. If we want to reduce the size more substantially, we need to review other dependencies. We should check how extensively we use a specific dependency, and if it's not used extensively, we should consider dropping it.


nandorsoma:
Fixes #581

PR #660, opened 2024-06-03
Profile enhancements

henneberger:


profiles/default/package.json
```diff
@@ -16,7 +16,7 @@
           "connector" : "kafka",
           "format" : "flexible-json",
           "properties.bootstrap.servers": "${PROPERTIES_BOOTSTRAP_SERVERS}",
-          "properties.group.id": "groupid",
+          "properties.group.id": "${PROPERTIES_GROUP_ID}",
           "scan.startup.mode" : "earliest-offset",
```
mbroecheler, line 20: this should be `group-offsets` (or omitted since that is the default). Otherwise, it will always start from the beginning.

profiles/default/kafka/Dockerfile.ftl
```diff
@@ -2,6 +2,11 @@ FROM docker.io/bitnami/kafka:3.4.0-debian-11-r38
 
 RUN echo '#!/bin/bash' > /opt/bitnami/scripts/kafka/create-topics.sh \
 <#list kafka["topics"] as topic>
-    && echo '/opt/bitnami/kafka/bin/kafka-topics.sh --create --bootstrap-server kafka:9092 --topic ${topic.name} --partitions ${topic.numPartitions} --replication-factor ${topic.replicationFactor}' >> /opt/bitnami/scripts/kafka/create-topics.sh \
+    && echo '/opt/bitnami/kafka/bin/kafka-topics.sh --create --if-not-exists --bootstrap-server kafka:9092 --topic ${topic.name} --partitions ${topic.numPartitions} --replication-factor ${topic.replicationFactor}' >> /opt/bitnami/scripts/kafka/create-topics.sh \
 </#list>
+<#if config["values"]?? && config["values"]["addl-topics"]??>
```
mbroecheler, line 7: Can we please rename this option to `create-topics`? It's a little more obvious imho.

mbroecheler (changes requested):
Two minor change requests. Thank you for getting this out quickly.

PR #662, opened 2024-06-04
Add bearer token handling to the graphiql endpoint

### LINKED ISSUE #661: Add bearer token handling to the graphiql endpoint
When `JWTAuthOptions` is configured on the Vertx server, automatically add token handling to the GraphiQL endpoints.

nandorsoma:
closes #661

### DIFF
--- a/sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/GraphQLServer.java
+++ b/sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/GraphQLServer.java
@@ -38,7 +38,6 @@ import io.vertx.core.Vertx;
 import io.vertx.core.http.HttpMethod;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.auth.jwt.JWTAuth;
-import io.vertx.ext.auth.jwt.JWTAuthOptions;
 import io.vertx.ext.healthchecks.HealthCheckHandler;
 import io.vertx.ext.web.Route;
 import io.vertx.ext.web.Router;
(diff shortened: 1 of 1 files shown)


sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/GraphQLServer.java
```diff
@@ -195,17 +198,41 @@ private void logRequests(Router router) {
...
+          "{ \"statusCode\": 500, \"statusMessage\": \"Internal Server Error\" }"
+      );
+    });
+    router.errorHandler(401, ctx -> {
+      log.error(String.format(
+          "Unauthorized. request.uri: [%s]",
+          ctx.request().uri()
+      ), ctx.failure());
+      ctx.response().setStatusCode(401).end(
+          "{ \"statusCode\": 401, \"statusMessage\": \"Unauthorized\" }"
```
nandorsoma, line 234: Normally, a JSON body as a response is not required when the status code is 401. However, GraphiQL does require it, which is why we're seeing the unusual error message on the UI. This message is a result of the JS code being unable to parse the response body...

sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/GraphQLServer.java
```diff
@@ -222,14 +249,14 @@ private void addHealthCheckHandler(Router router) {
   }
 
   private Optional<JWTAuthHandler> createJwtAuthHandler() {
-    if (this.config.getJWTAuthOptions() != null) {
-      JWTAuth provider = JWTAuth.create(vertx, config.getJWTAuthOptions());
-      return Optional.of(JWTAuthHandler.create(provider));
+    if (config.getJWTAuthOptions() != null) {
+      authProvider = JWTAuth.create(vertx, config.getJWTAuthOptions());
```
nandorsoma, line 253: I know it is not a priority, but this is a good example why using a dependency injection framework could be beneficial. Could we consider adopting one in the long term?

PR #667, opened 2024-06-09
Add auth to vertx graphql

henneberger:


mbroecheler (approved):
Love the test case - thank you!

PR #706, opened 2024-07-05
Implementing Logging Functionality

nandorsoma:
This PR introduces logging functionality.

To enable logging, we must use the `EXPORT TableName TO log.TableName` statement.

The behavior of the compiler varies based on the value of `compiler.log`. The possible values are:

- `none`: Logging is disabled, and no connector for logging will be generated.
- `print`: The print connector will be used, similar to when we execute `EXPORT TableName TO print.TableName`.
- `kafka`: A Kafka connector will be created, along with a table-specific log topic in Kafka.

Example package.json for kafka:
```json
{
  "version": "1",
  "enabled-engines": ["kafka", "flink"],
  "compiler": {
    "log": "kafka"
  }
}
```


nandorsoma:
Opened against wrong branch.

PR #715, opened 2024-07-09
Implementing Logging Functionality

nandorsoma:
Original PR: https://github.com/DataSQRL/sqrl/pull/707
Due to the complexity of resolving conflicts in the original PR, I've created a new one. This new PR includes a single commit based on a rebase from version 0.5, along with additional commits aimed at fixing errors that weren't addressed during the conflict resolution.

This PR introduces logging functionality.

To enable logging, we must use the `EXPORT TableName TO logger.TableName` statement.

The behavior of the compiler varies based on the value of `compiler.log`. The possible values are:

- `none`: Logging is disabled, and no connector for logging will be generated.
- `print`: The print connector will be used, similar to when we execute `EXPORT TableName TO print.TableName`.
- `kafka`: A Kafka connector will be created, along with a table-specific log topic in Kafka.

Example package.json for kafka:
```json
{
  "version": "1",
  "enabled-engines": ["kafka", "flink"],
  "compiler": {
    "logger": "kafka"
  }
}
```


sqrl-base/src/main/java/com/datasqrl/config/SystemBuiltInConnectors.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.Getter;
+
+@AllArgsConstructor
+@Getter
+public
+enum SystemBuiltInConnectors {
+
+  PRINT_SINK(Name.system("print"), ExternalDataType.sink),
+  LOCAL_FILE_SOURCE(Name.system("localfile"), ExternalDataType.source),
+  LOG(Name.system("log"), ExternalDataType.sink);
```
henneberger, line 16: im not sure why we need this... We need to book keep very little when a 'log' export is done.

sqrl-calcite/src/main/java/com/datasqrl/plan/validate/ScriptPlanner.java
```diff
@@ -221,40 +229,64 @@ public Void visit(SqrlImportDefinition node, Void context) {
...
+    ModifiableTable modTable = table.get().unwrap(ModifiableTable.class);
+
+    int numSelects = modTable.getNumSelects();
+    RelNode exportRelNode = planner.getRelBuilder().scan(modTable.getNameId()).build();
+
+    Optional<SystemBuiltInConnectors> builtInSink = SystemBuiltInConnectors.forExport(sinkPath.popLast());
+    TableSink tableSink = null;
+    if (builtInSink.isPresent()) {
+      SystemBuiltInConnectors connector = builtInSink.get();
+      if (connector == SystemBuiltInConnectors.LOG) {
```
henneberger, line 246: this entire if block doesn't seem necessary. They are provided by the connector factory. The connector factory could bookkeep easily since it is aware of the type of sink its creating

sqrl-planner/src/main/java/com/datasqrl/loaders/ObjectLoaderImpl.java
```diff
@@ -45,15 +47,19 @@ public class ObjectLoaderImpl implements ObjectLoader {
...
       CalciteTableFactory tableFactory, ModuleLoader moduleLoader,
-      TableConfigLoader tableConfigFactory) {
+      TableConfigLoader tableConfigFactory, PackageJson sqrlConfig, LogManager logManager) {
     this.resourceResolver = resourceResolver;
     this.errors = errors;
     this.tableFactory = tableFactory;
     this.moduleLoader = moduleLoader;
     this.tableConfigFactory = tableConfigFactory;
+    this.sqrlConfig = sqrlConfig;
+    this.logManager = logManager;
```
henneberger, line 62: object loader is just for loading modules. it shouldn't be aware of a log manager?

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/ConnectorConfigImpl.java
```diff
@@ -26,7 +26,8 @@ public class ConnectorConfigImpl implements TableConfig.ConnectorConfig {
       "filesystem", TableType.STREAM,
       "upsert-kafka", TableType.VERSIONED_STATE,
       "jdbc", TableType.LOOKUP,
-      "jdbc-sqrl", TableType.LOOKUP
+      "jdbc-sqrl", TableType.LOOKUP,
+      "datagen", TableType.STREAM
```
henneberger, line 30: anything not identified is a stream by default

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/ConnectorFactoryFactoryImpl.java
```diff
@@ -16,31 +16,53 @@
...
+      default:
+        throw new IllegalArgumentException("Unknown connector: " + builtInConnector);
+    }
+  }
+
   @Override
-  public Optional<ConnectorFactory> create(Type type, String name) {
+  public Optional<ConnectorFactory> create(Type engineType, String connectorName) {
+
+    // from conflict, include into the new logic
```
henneberger, line 44: ?

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/ConnectorFactoryFactoryImpl.java
```diff
@@ -148,4 +170,14 @@ private ConnectorFactory createJdbcConnectorFactory(ConnectorConf connectorConf)
       return builder.build();
     };
   }
+
+  private ConnectorFactory createBlackHoleConnectorFactory() {
```
henneberger, line 174: used?

henneberger (commented):
To answer you question about where DI happens: We don't currently have it threaded into the early init part of datasqrl. We should bring it in eventually to make it cohesive but theres a bit of rearchitecture to make that happen

sqrl-base/src/main/java/com/datasqrl/config/PackageJson.java
```diff
@@ -49,6 +49,8 @@ interface CompilerConfig {
     void setSnapshotPath(String string);
 
     boolean isAddArguments();
+
+    LogMethod getLog();
```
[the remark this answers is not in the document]
nandorsoma: Originally `kafka` was `log-engine`/`engine` but @mbroecheler suggested that from the user point of view it is easier to understand.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/ConnectorFactoryFactoryImpl.java
nandorsoma: No, that remained from an intermediate solution where I routed LogMethod.NONEs to blackhole.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/ConnectorFactoryFactoryImpl.java
nandorsoma: Prior to both PRs, this section was somewhat disorganized, allowing us to create connectors based on names and engineType. The logging PR introduced some refactoring, making engineType the sole determinant for selecting a factory. However, in the snowflake PR, you used the connectorName to decide whether to create an iceberg connector. I considered whether we could use the engineType for this purpose, but I'm not certain. Given the workaround comment, I decided it wasn't worth investigating at this point. I merely wanted to highlight that the inconsistency in the logic arose from a conflict, resulting in a state that might benefit from further refactoring.

sqrl-base/src/main/java/com/datasqrl/config/SystemBuiltInConnectors.java
nandorsoma: This section was added by @mbroecheler, so he would be the best person to explain its purpose. From my understanding, it aids comprehension. It seems unusual that we check whether we can create a ConnectorFactory in the ImportExportAnalyzer without using it. Later, in the ScriptPlanner, we create the same ConnectorFactory and then use it. I believe it's cleaner to check here if we have a built-in connector and then later create the corresponding connector definition for it.

sqrl-calcite/src/main/java/com/datasqrl/plan/validate/ScriptPlanner.java
nandorsoma: Do you mean bookkeeping the logs? This part is about using the LogManager to bookkeep the logs which is probably more suited for that task than a factory.

sqrl-planner/src/main/java/com/datasqrl/loaders/ObjectLoaderImpl.java
nandorsoma: Yes, I agree in the previous PR it was better when we just injected the LogManager into the ScriptPlaner. However since that object is created manually I don't see any other option to inject these classes into it.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/ConnectorConfigImpl.java
nandorsoma: Wouldn't it be more beneficial to be explicit in this case? If not, why do we specify filesystem and Kafka? I was simply trying to follow an existing pattern. Additionally, I remember encountering an error when the connector wasn't specified.

sqrl-base/src/main/java/com/datasqrl/config/SystemBuiltInConnectors.java
mbroecheler: These are the connectors that are "hard-coded" into SQRL and always available. They are instantiated based on configuration in the package.json.
The goal was to have these centrally defined so that we do not have string constants floating all over the codebase. This makes it easier to understand what the pre-defined connectors are, what their name is, and whether they are sources or sinks.
I added this mostly to refactor what we had before because I noticed that we had a proliferation of string constants.

sqrl-base/src/main/java/com/datasqrl/config/PackageJson.java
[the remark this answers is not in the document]
mbroecheler: The user needs to be able to configure the logging independently of the log engine. There are 3 options currently:
- none: don't do any logging
- print: log to stdout
- name of the log engine: log to the log engine

sqrl-base/src/main/java/com/datasqrl/config/PackageJson.java
```diff
@@ -125,4 +127,29 @@ interface DataDiscoveryConfig {
 
     ErrorCollector getErrors();
   }
+
+  enum LogMethod {
```
mbroecheler, line 131: This does not look correct from a product perspective: we are hard-coding what sinks we allow for the logging here. That should be resolved dynamically from the list of engines that are configured and the list of systembuildinconnectors. We don't want to duplicate that information imho.

mbroecheler: More specifically: when we add a new log engine (e.g. postgres-cdc) we should not have to go in here and add another value. That breaks modularity.

sqrl-planner/src/main/java/com/datasqrl/loaders/ObjectLoaderImpl.java
mbroecheler: @henneberger Do you have a better idea for the modularity boundary here? I'm not a big fan of the PackageJson and the LogManager being in the ScriptPlanner. In particular now that it cannot be injected anymore.
Should the LogManager live in the Schema (since it's essentially a schema element)?
Where should the configuration live so the ScriptPlanner has access to it?

mbroecheler (changes requested):
Looks mostly good. Thank you! See comments inline and slack message for more details.

sqrl-base/src/main/java/com/datasqrl/config/PackageJson.java
nandorsoma: I understand your point. However, I also see the benefit in enumerating the available configuration options. This way, if you need to quickly determine the available options, you can simply refer to one class file. If it becomes dynamic, the code maintainer would need to delve deeper into the code just to answer this question. Nonetheless, I see the value in both approaches. I've made the changes as you requested. Please let me know if it's satisfactory!

mbroecheler (approved):

sqrl-base/src/main/java/com/datasqrl/config/PackageJson.java
mbroecheler: thanks for changing that.

henneberger (approved):

PR #718, opened 2024-07-10
Add json functions for snowflake

henneberger:


nandorsoma (approved):
LGTM, one thing I noticed that currently the test's package.json is pointing to a custom s3 location. I wanted to test it manually, but it requires extra setup which made me think whether we want to write an IT test for it which utilizes https://github.com/adobe/S3Mock testcontainer.

Issue #731, opened 2024-07-26
Add support to import from and export to the log engine


nandorsoma:
Make it possible to export to and import from the log engine. This way we could convert a state table to a stream:

```
MerchantState := DISTINCT Merchant ON merchantId ORDER BY updatedTime DESC; 

EXPORT MerchantState TO log.MerchantLog;
IMPORT log.MerchantLog;
```

PR #747, opened 2024-08-06
Add support to import from and export to the log engine

nandorsoma:
#731


sqrl-calcite/src/main/java/com/datasqrl/plan/validate/ScriptPlanner.java
```diff
@@ -210,24 +210,30 @@ public Void visit(SqrlImportDefinition node, Void context) {
     }
 
     NamePath path = nameUtil.toNamePath(node.getImportPath().names);
-    SqrlModule module = moduleLoader.getModule(path.popLast()).orElse(null);
-
-    if (module == null) {
-      throw addError(ErrorCode.GENERIC, node, "Could not find module [%s] at path: [%s]",
-          path, String.join("/", path.toStringList()));
-    }
+    if (path.getFirst().getDisplay().equals("log")) {
```
henneberger, line 213: we can move this into the module loader as a custom module

henneberger: Added #785 to address

PR #771, opened 2024-08-24
Iceberg Modularity and Partition Fixes

mbroecheler:
This PR does two things:
1) strives to find a better architectural separation between the table format engine (i.e. iceberg) and the associated query engines (i.e. duckdb or snowflake) by allowing multiple query engines to be registered with a single table format and properly delegating from the table format engine to the query engine.

2) fixes the partition assignment in Flink by accounting for the offset (since the columns after the offset are for sorting only).

NOTE: This PR makes changes to the iceberg profile files.


profiles/default/iceberg/snowflake-views.sql.ftl
```diff
@@ -0,0 +1,6 @@
+<#if iceberg?? && iceberg["engines"]["snowflake"]??>
+<#list iceberg["engines"]["snowflake"]["views"] as statement>
+${statement["sql"]};
```
henneberger, line 3: I'll move this to the snowflake engine, it makes more sense for it to be there. This should have schema for iceberg related settings, like registering schemas in aws glue et al. I'll make the change.

sqrl-planner/src/main/java/com/datasqrl/engine/database/DatabasePhysicalPlan.java
```diff
@@ -4,13 +4,23 @@
...
+import com.datasqrl.sql.SqlDDLStatement;
+import java.util.List;
 import java.util.Map;
 
 public interface DatabasePhysicalPlan extends EnginePhysicalPlan {
 
+  List<SqlDDLStatement> getDdl();
+
+  default boolean removeIndexDdl() {
+    return getDdl().removeIf(ddl -> (ddl instanceof CreateIndexDDL) || (ddl instanceof DropIndexDDL));
```
henneberger, line 21: i think the root issue is that the engine also delefates to create 'normal' base jdbc tables. It is unnecessary in duckdb as no schema needs to be created, so no DDL indices should be created at all.

sqrl-planner/src/main/java/com/datasqrl/engine/database/QueryEngine.java
```diff
@@ -13,9 +20,14 @@
  */
 public interface QueryEngine extends ExecutionEngine {
 
-
-  default Map<IdentifiedQuery, QueryTemplate> updateQueries(
-      ConnectorFactoryFactory connectorFactory, EngineConfig connectorConfig, Map<IdentifiedQuery, QueryTemplate> queries) {
-    return queries;
+  @Override
+  default DatabasePhysicalPlan plan(StagePlan plan, List<StageSink> inputs,
```
henneberger, line 24: This 'updateQueries' is a bit of an antipattern for the engine planning. We can safely just delegate the queries to the engine and allow it to part of its own physical plan, which lets this be removed entirely. This is my fault and I'll correct it on top of this PR.

sqrl-planner/src/main/java/com/datasqrl/engine/stream/flink/plan/SqrlToFlinkSqlGenerator.java
```diff
@@ -379,23 +380,22 @@ private SqlCreateTable registerSinkTable(WriteSink sink, RelNode relNode,
...
-          String tableId = engineSink.getNameId();
-          Optional<IndexDefinition> optIndex =  dbPlan.getIndexDefinitions().stream()
-              .filter(i -> i.getTableId().equals(tableId))
-              .findFirst();
-          if (optIndex.isPresent()) {
-            IndexDefinition mainIndex = optIndex.get();
-            builder.setPartitionKey(mainIndex.getColumnNames());
-          }
-
+      if (engineSink.getStage().supportsFeature(EngineFeature.PARTITIONING)) {
```
henneberger, line 388: 👍

henneberger: I'm going to add a configuration switch for partition and primary keys so we can target all permutations of the iceberg connector. This code shouldn't change, we can move it to the engine.

sqrl-planner/src/main/java/com/datasqrl/engine/stream/flink/plan/SqrlToFlinkSqlGenerator.java
```diff
@@ -379,23 +380,22 @@ private SqlCreateTable registerSinkTable(WriteSink sink, RelNode relNode,
...
+        DatabaseStagePlan dbPlan = (DatabaseStagePlan) stagePlan;
+        String tableId = engineSink.getNameId();
+        Optional<IndexDefinition> optIndex =  dbPlan.getIndexDefinitions().stream()
+            .filter(idx -> idx.getTableId().equals(tableId))
+            .filter(idx -> idx.getType().isPartitioned())
+            .findFirst();
+        if (optIndex.isPresent()) {
+          IndexDefinition partitionIndex = optIndex.get();
+          List<String> partitionColumns = partitionIndex.getColumnNames().subList(0, partitionIndex.getPartitionOffset());
```
henneberger, line 397: Huh, this could be deceptive from an interface perspective.

henneberger:
I can make the changes i outlined, let me know if you're comfortable with this and I can land this and make those changes.

profiles/default/iceberg/snowflake-views.sql.ftl
mbroecheler: Agreed, each query engine should be it's own folder so we can put the DDL (if any) and view definitions there.

sqrl-planner/src/main/java/com/datasqrl/engine/database/DatabasePhysicalPlan.java
mbroecheler: Yes, this is a bit of a hack. IMHO the root is that "indexes" don't exist for table formats and associated query engines (you can only partition data but not index yet), yet indexes are important for normal databases, which is the foundation of the JDBCEngine.

I think we should have a "BaseJDBCEngine" which only does table creation and then extend it to a OperationalJDBCEngine (i.e. postgres, mysql, sqlserver, etc) that supports indexes and subscribe/notify and AnalyticJDBCEngine which does not. That would be a cleaner seperation.

sqrl-planner/src/main/java/com/datasqrl/engine/stream/flink/plan/SqrlToFlinkSqlGenerator.java
mbroecheler: What would be a way to make this more obvious? 
A partitioned btree is defined by a set of columns for the partition followed by a set of columns for the sort order.
In many systems (e.g. Cassandra, Citrix) that's indicated by an offset.

PR #827, opened 2024-09-21
JSON Schema Validation

### LINKED ISSUE #826: JSON Schema validation
Table.json + package.json

mbroecheler:
for table.config.json and package.json

Fixes #826


sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfigCommons.java
```diff
@@ -321,49 +344,83 @@ public boolean hasKey(String key) {
...
-  public static TableConfigImpl fromFilesTableConfig(@NonNull Name name, ErrorCollector errors, @NonNull List<Path> files) {
-    return new TableConfigImpl(name, fromFiles(errors, files));
+  public static TableConfigImpl fromFilesTableConfig(@NonNull Name name, ErrorCollector errors,
+      @NonNull List<Path> files) {
+    return new TableConfigImpl(name, fromFiles(errors, Optional.of("/jsonSchema/tableConfig.json"), files));
   }
 
   public static PackageJson fromFilesPackageJson(ErrorCollector errors, @NonNull List<Path> files) {
-    return new PackageJsonImpl(fromFiles(errors, files));
+    return new PackageJsonImpl(fromFiles(errors, Optional.empty(), files));
```
mbroecheler, line 358: @nandorsoma Can you please add a json schema for package.json files and add it here?

mbroecheler:
@nandorsoma Can you please take this PR over from me and do the following:
1) review it
2) add validation for package.json files by following the same pattern as for the tableconfig files (see comment I added in the code). OpenAI is pretty good at generating the json schema files. I just gave it an example and told it what the rules are.
3) add more test cases (for package.json validation) in ConfigurationSchemaTest

pom.xml
```diff
@@ -698,6 +699,12 @@
         <artifactId>jackson-dataformat-yaml</artifactId>
         <version>${jackson.version}</version>
       </dependency>
+      <!-- JSON SChema -->
+      <dependency>
+        <groupId>com.networknt</groupId>
```
nandorsoma, line 704: VertX has a nice library to handle schema validation `io.vertx:vertx-json-schema`. Since we already have a bunch of vertx dependencies, maybe it would be beneficial to use it instead of pulling a dependency from a "different family". I don't feel strongly though. What do you think?

nandorsoma: Quickly looked at it and maybe the vertx library is too loaded for this. Not sure for eg. for what vertx-core is used.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfigCommons.java
```diff
@@ -321,49 +344,83 @@ public boolean hasKey(String key) {
...
+    SqrlConfigCommons newconfig = new SqrlConfigCommons(errors, null, config, "");
     newconfig.setProperty(VERSION_KEY, version);
     return newconfig;
   }
 
-  public static TableConfigImpl fromFilesTableConfig(@NonNull Name name, ErrorCollector errors, @NonNull List<Path> files) {
-    return new TableConfigImpl(name, fromFiles(errors, files));
+  public static TableConfigImpl fromFilesTableConfig(@NonNull Name name, ErrorCollector errors,
+      @NonNull List<Path> files) {
+    return new TableConfigImpl(name, fromFiles(errors, Optional.of("/jsonSchema/tableConfig.json"), files));
```
nandorsoma, line 354: What's is the added value of using Optionals here? Couldn't we just pass the raw string here?

mbroecheler: Yes, let's do raw string

pom.xml
mbroecheler: I was looking at that. We don't depend on vertx in the main code line, only for the server. Hence I thought this library, based on jackson, seemed like a better fit.

nandorsoma (approved):

nandorsoma:
Unfortunately the commits are not signed, therefore I'm not allowed to merge it. I believe you have the right @mbroecheler to do so.

PR #836, opened 2024-09-26
Move to old run command, flink 1.19 deps

henneberger:


sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/cmd/AbstractCompilerCommand.java
```diff
@@ -83,9 +86,12 @@ public void execute(ErrorCollector errors, String[] profiles, Path snapshotPath,
     }
 
     if (goal == ExecutionGoal.TEST) {
-      List<String> defaultKeys = List.of(EngineKeys.TEST, EngineKeys.DATABASE, EngineKeys.LOG,
-          EngineKeys.SERVER, EngineKeys.STREAMS);
-      sqrlConfig.setPipeline(defaultKeys);
+      Set<String> enabledEngines = new HashSet<>(sqrlConfig.getEnabledEngines());
+      enabledEngines.add(EngineKeys.TEST);
```
mbroecheler, line 90: This seems problematic. The test command only supports the engines that we have explicitly included. If the user has configured some other engine it will fail.

mbroecheler (changes requested):
Skipped the flink-jdbc part assuming it's mostly copy-paste.
Some minor nits and comments. Thank you!

sqrl-testing/sqrl-integration-tests/pom.xml
```diff
@@ -130,5 +136,340 @@
...
+      <scope>compile</scope>
+      <classifier>test</classifier>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-streams</artifactId>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
```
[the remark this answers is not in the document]
henneberger: the classifier is required, not sure theres a way around this

sqrl-tools/sqrl-test/src/main/java/com/datasqrl/SnapshotOkException.java
```diff
@@ -0,0 +1,14 @@
+package com.datasqrl;
+
+public class SnapshotOkException extends Exception {
+  private final String testName;
+
+  public SnapshotOkException(String testName) {
+    super("Snapshot created for test: " + testName + ". Rerun to verify.");
```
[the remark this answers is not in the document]
henneberger: Ah you caught it. It's a bit of a hack so we display errors and passing snapshots in the same place :). We can fix it.

sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/cmd/AbstractCompilerCommand.java
henneberger: The problem is that then the sqrl script can't be compiled because it's missing engines (if annotated with iceberg for example).

mbroecheler: Let's move that to a separate ticket then

mbroecheler (approved):

PR #850, opened 2024-10-02
Add graphql subscription support for postgres-log engine in vertx

nandorsoma:
This is a draft pull request. There are still some tasks to complete, which I have marked with TODOs.

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/ddl/statements/notify/CreateNotifyTriggerDDL.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/ddl/statements/notify/CreateNotifyTriggerDDL.java
@@ -4,7 +4,6 @@ import com.datasqrl.sql.SqlDDLStatement;
 import com.google.common.base.Preconditions;
 import java.util.List;
 import java.util.stream.Collectors;
-import lombok.AllArgsConstructor;
 import lombok.NonNull;
 
 public class CreateNotifyTriggerDDL implements SqlDDLStatement {
--- a/sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/ddl/statements/notify/CreateNotifyTriggerDDL.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/ddl/statements/notify/CreateNotifyTriggerDDL.java
@@ -37,7 +37,7 @@ public class CreateNotifyTriggerDDL implements SqlDDLStatement {
   private String createPayload() {
     String argumentList = primaryKeys.stream()
         .map(pk ->
-            String.format("'%s', NEW.%s", pk, pk))
+            String.format("'%s', NEW.\"%s\"", pk, pk))
         .collect(Collectors.joining(", "));
 
     return String.format("jsonb_build_object(%s)::text", argumentList);
--- a/sqrl-planner/src/main/java/com/datasqrl/graphql/inference/GraphqlModelGenerator.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/graphql/inference/GraphqlModelGenerator.java
@@ -13,6 +13,7 @@ import com.datasqrl.config.TableConfig;
 import com.datasqrl.engine.PhysicalPlan.StagePlan;
 import com.datasqrl.engine.database.QueryTemplate;
 import com.datasqrl.engine.database.relational.ddl.statements.InsertStatement;
+import com.datasqrl.engine.database.relational.ddl.statements.notify.ListenNotifyAssets;
 import com.datasqrl.engine.log.Log;
 import com.datasqrl.engine.log.kafka.KafkaPhysicalPlan;
 import com.datasqrl.engine.log.postgres.PostgresLogPhysicalPlan;
@@ -107,7 +108,20 @@ public class GraphqlModelGenerator extends SchemaWalker {
       subscriptionCoords = new KafkaSubscriptionCoords(fieldName, topic, Map.of(), filters);
     } else if (logPlan.isPresent() && logPlan.get() instanceof PostgresLogPhysicalPlan) {
       String tableName = (String) log.getConnectorContext().getMap().get("table-name");
-      subscriptionCoords = new PostgresSubscriptionCoords(fieldName, tableName, filters);
+
+      ListenNotifyAssets listenNotifyAssets = ((PostgresLogPhysicalPlan) logPlan.get())
+          .getQueries().stream()
+          .filter(query -> query.getListen().getTableName().equals(tableName))
+          .findFirst()
+          .orElseThrow(
+              () -> new RuntimeException("Could not find query statement for table: " + tableName)
+          );
+
+      subscriptionCoords = new PostgresSubscriptionCoords(
+          fieldName, tableName, filters,
+          listenNotifyAssets.getListen().getSql(),
+          listenNotifyAssets.getOnNotify().getSql(),
+          listenNotifyAssets.getParameters());
     } else {
       throw new RuntimeException("Unknown log plan: " + logPlan.getClass().getName());
     }
--- a/sqrl-planner/src/main/java/com/datasqrl/graphql/inference/GraphqlModelGenerator.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/graphql/inference/GraphqlModelGenerator.java
@@ -25,11 +25,13 @@ import com.datasqrl.graphql.server.RootGraphqlModel.DuckDbQuery;
 import com.datasqrl.graphql.server.RootGraphqlModel.FieldLookupCoords;
 import com.datasqrl.graphql.server.RootGraphqlModel.JdbcQuery;
 import com.datasqrl.graphql.server.RootGraphqlModel.KafkaMutationCoords;
+import com.datasqrl.graphql.server.RootGraphqlModel.KafkaSubscriptionCoords;
 import com.datasqrl.graphql.server.RootGraphqlModel.MutationCoords;
 import com.datasqrl.graphql.server.RootGraphqlModel.PagedDuckDbQuery;
 import com.datasqrl.graphql.server.RootGraphqlModel.PagedJdbcQuery;
 import com.datasqrl.graphql.server.RootGraphqlModel.PostgresLogMutationCoords;
 import com.datasqrl.graphql.server.RootGraphqlModel.PagedSnowflakeDbQuery;
+import com.datasqrl.graphql.server.RootGraphqlModel.PostgresSubscriptionCoords;
 import com.datasqrl.graphql.server.RootGraphqlModel.SnowflakeDbQuery;
 import com.datasqrl.graphql.server.RootGraphqlModel.SubscriptionCoords;
 import com.datasqrl.io.tables.TableSource;
@@ -94,9 +96,23 @@ public class GraphqlModelGenerator extends SchemaWalker {
       filters.put(input.getName(), field.getName());
     }
 
-    subscriptions.add(new SubscriptionCoords(fieldDefinition.getName(),
-        (String)log.getConnectorContext().getMap().get("topic"), Map.of(),
-        filters));
+    Optional<EnginePhysicalPlan> logPlan = getLogPlan();
+
+    String fieldName = fieldDefinition.getName();
+
+    SubscriptionCoords subscriptionCoords;
+
+    if (logPlan.isPresent() && logPlan.get() instanceof KafkaPhysicalPlan) {
+      String topic = (String) log.getConnectorContext().getMap().get("topic");
+      subscriptionCoords = new KafkaSubscriptionCoords(fieldName, topic, Map.of(), filters);
+    } else if (logPlan.isPresent() && logPlan.get() instanceof PostgresLogPhysicalPlan) {
+      String tableName = (String) log.getConnectorContext().getMap().get("table-name");
+      subscriptionCoords = new PostgresSubscriptionCoords(fieldName, tableName, filters);
+    } else {
+      throw new RuntimeException("Unknown log plan: " + logPlan.getClass().getName());
+    }
+
+    subscriptions.add(subscriptionCoords);
   }
 
   @Override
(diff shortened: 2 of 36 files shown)


sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/ddl/statements/notify/CreateNotifyTriggerDDL.java
```diff
@@ -37,7 +37,7 @@ public String getSql() {
   private String createPayload() {
     String argumentList = primaryKeys.stream()
         .map(pk ->
-            String.format("'%s', NEW.%s", pk, pk))
+            String.format("'%s', NEW.\"%s\"", pk, pk))
```
henneberger, line 40: using calcite ast tokens can handle these nuances for us in a consistent way across all query generation, these little bugs can be annoying to run into

sqrl-planner/src/main/java/com/datasqrl/graphql/inference/GraphqlModelGenerator.java
```diff
@@ -158,6 +187,15 @@ protected void walkMutation(APISource source, TypeDefinitionRegistry registry,
     mutations.add(mutationCoords);
   }
 
+  private Optional<EnginePhysicalPlan> getLogPlan() {
+    //Todo: Bad instance checking, Fix to bring multiple log engines (?)
```
henneberger, line 191: there are a lot of assumptions of a single log engine in the vertx server too. we can fix this down the road when we introduce calcite as a query delegation layer.

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/server/RootGraphqlModel.java
```diff
@@ -153,15 +152,63 @@ public DataFetcher<?> accept(MutationCoordsVisitor visitor) {
...
+      use = JsonTypeInfo.Id.NAME,
+      property = "type"
+  )
+  @JsonSubTypes({
+      @Type(value = KafkaSubscriptionCoords.class, name = "kafka"),
+      @Type(value = PostgresSubscriptionCoords.class, name = "postgres_log")
+  })
+  public static abstract class SubscriptionCoords {
+    protected String type;
+    public abstract DataFetcher<?> accept(SubscriptionCoordsVisitor visitor);
```
henneberger, line 165: Use `<R, C> R accept(SubscriptionCoordsVisitor<R, C> visitor, C context)`

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/server/GraphQLEngineBuilder.java
```diff
@@ -110,9 +110,9 @@ public GraphQL.Builder visitRoot(RootGraphqlModel root, Context context) {
 
     if (root.subscriptions != null) {
       for (SubscriptionCoords sc : root.subscriptions) {
+        DataFetcher<?> subscriptionFetcher = sc.accept(context.createSubscriptionFetcherVisitor());
```
henneberger, line 113: It might be cleaner to do pass a mutation/subscription factory into GraphQLEngineBuilder rather than shuttle it through the context. I was using the context as a way to keep jdbc connection details and using it also for object construction was probably a mistake on my part.

sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/GraphQLServer.java
```diff
@@ -356,32 +347,47 @@ private SqlClient getPostgresSqlClient() {
...
+      Promise<Void> startPromise, VertxJdbcClient client) {
     Map<String, SinkConsumer> consumers = new HashMap<>();
     for (SubscriptionCoords sub: root.getSubscriptions()) {
-      KafkaConsumer<String, String> consumer = KafkaConsumer.create(vertx, getSourceConfig());
-      consumer.subscribe(sub.getTopic())
-          .onSuccess(v -> log.info("Subscribed to topic: {}", sub.getTopic()))
-          .onFailure(startPromise::fail);
-
-      consumers.put(sub.getFieldName(), new KafkaSinkConsumer<>(consumer));
+      if (sub instanceof KafkaSubscriptionCoords) {
```
henneberger, line 371: Reuse the same visitor pattern instead of instance checking?

sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/postgres_log/PostgresListenNotifyConsumer.java
```diff
@@ -0,0 +1,111 @@
...
+        // Execute the LISTEN command to listen for notifications on a specific channel
+        pgConnection.query(listenQuery).execute(queryResult -> {
+          if (queryResult.succeeded()) {
+            log.info("LISTEN command executed successfully: {}", listenQuery);
+          } else {
+            log.error(String.format("Unable to execute LISTEN command: %s", queryResult.cause().getMessage()), queryResult.cause());
+          }
+        });
+      } else {
+        log.error(String.format("Unable to establish connection. %s", res.cause().getMessage()), res.cause());
```
henneberger, line 58: possible to fail the future?

sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/postgres_log/PostgresListenNotifyConsumer.java
```diff
@@ -0,0 +1,111 @@
...
+
+    preparedQuery.execute(Tuple.from(paramObj))
+        .onComplete(asyncResult -> {
+          if (asyncResult.succeeded()) {
+            RowSet<Row> rows = asyncResult.result();
+            for (Row row : rows) {
+              Map<String, Object> resultMap = new HashMap<>();
+              for (Row rowValue : rows) {
+                for (int i = 0; i < row.size(); i++) {
+                  resultMap.put(row.getColumnName(i), rowValue.getValue(i));
```
henneberger, line 95: rowValue.toJson() then pass the JsonObject to the listener (see resultMapper)

sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/postgres_log/PostgresListenNotifyConsumer.java
```diff
@@ -0,0 +1,111 @@
...
+        paramObj.add(jsonPayload.getInteger(parameter));
+      } else {
+        throw new IllegalArgumentException("Unknown parameter: " + parameter);
+      }
+    }
+
+    // TODO (Soma) - It feels odd that we are using vertx a bit differently compared to how we
+    //  handle the notifications. This is the accepted way of running queries in the codebase
+    //  however in case on notifications we are forced to use PGConnection since that's the
+    //  only way currently to listen to notifications.
```
henneberger, line 84: This feels fine. We have an explicit connection to listen to notifications (only would be a few connections since its shared), and then use the connection pooler to do the queries. 

So are we executing these queries regardless if we have a listener? Does that make sense to do?

nandorsoma: > This feels fine. We have an explicit connection to listen to notifications (only would be a few connections since its shared), and then use the connection pooler to do the queries.
Ok.

> So are we executing these queries regardless if we have a listener? Does that make sense to do?
No, we are only executing the onNotify queries when a notification arrives. But I'm not sure what do you want to hint with this question. Am I missing something?

henneberger: High level there are two approaches:
- We have one database connection per subscription type and we use flux + calcite or trivial filtering logic to fan out all of the requests
- We create a new database connection per subscription _connection_. We push down filters into the database and/or use calcite/trivial filters.

I'm just following up with what approach we are using and if there will be queries against the database if there are no subscription connections.

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/ddl/statements/notify/CreateNotifyTriggerDDL.java
nandorsoma: As agreed with @mbroecheler I'm going to skip this now.

sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/postgres_log/PostgresListenNotifyConsumer.java
nandorsoma: Yes, why not?

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/ddl/statements/notify/CreateNotifyTriggerDDL.java
henneberger: 👍. Calcite AST classes look like this. It has an 'unparse' method where we can delegate how quoting is handled to the calcite framework. https://github.com/DataSQRL/sqrl/blob/bd70c7ec619cc5d442e001852f71f526e52f4ba7/sqrl-calcite/src/main/java/com/datasqrl/calcite/dialect/postgres/SqlCreatePostgresView.java#L57

sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/postgres_log/PostgresListenNotifyConsumer.java
nandorsoma: So I believe for both options we need to have at least 2 connections. We need to keep one connection open and dedicated to receiving notifications. You cannot reuse it for other operations while you are listening on it.

Apart from that, do I understand correctly that this topic is not a blocker on merging this PR?

sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/postgres_log/PostgresListenNotifyConsumer.java
```diff
@@ -65,16 +65,22 @@ public void handleNotification(PgNotification notification, Consumer<Object> lis
...
-      if (parameter.equals("timeSec")) {
-        paramObj.add(OffsetDateTime.parse(jsonPayload.getString(parameter)));
+      Object value = jsonPayload.getValue(parameter);
 
-      } else if (parameter.equals("sensorid")) {
-        paramObj.add(jsonPayload.getInteger(parameter));
+      if (value instanceof String) {
+        // Check if it's a datetime string and parse it
+        String strValue = (String) value;
+        if (strValue.matches("^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}.*$")) {
```
mbroecheler, line 77: Why does this require special handling for postgres @nandorsoma ? We are also consuming json data in the kafka listener what is the issue with reusing that code?

sqrl-server/sqrl-server-vertx/src/main/java/com/datasqrl/graphql/postgres_log/PostgresDataFetcherFactory.java
```diff
@@ -0,0 +1,63 @@
...
+          if (data instanceof Map) {
+            objectMap = (Map) data;
+          } else if (data instanceof JsonObject) {
+            objectMap = ((JsonObject)data).getMap();
+          } else {
+            objectMap = Map.of();
+          }
+
+          Object retrievedData = objectMap.get(filter.getValue());
+          // TODO (Soma) - This is needed because it seems like in graphql there is no Long type.
```
mbroecheler, line 45: Why is this issue specific to postgres?

mbroecheler (commented):
Added some questions

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/server/RootGraphqlModel.java
```diff
@@ -148,20 +147,68 @@ public PostgresLogMutationCoords(String fieldName, String tableName, String inse
...
+  )
+  @JsonSubTypes({
+      @Type(value = KafkaSubscriptionCoords.class, name = "kafka"),
+      @Type(value = PostgresSubscriptionCoords.class, name = "postgres_log")
+  })
+  public static abstract class SubscriptionCoords {
+    protected String type;
+    public abstract <R, C> R accept(SubscriptionCoordsVisitor<R, C> visitor, C context);
+    public abstract String getFieldName();
+  }
```
henneberger, line 167: This needs to be backwards compatible, if there is no type then we should default to kafka. We've had to deal with several issues with older configs not being backwards compatible.

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/server/GraphQLEngineBuilder.java
```diff
@@ -70,12 +71,35 @@ public class GraphQLEngineBuilder implements
...
+
+    public Builder withMutationConfiguration(MutationConfiguration<DataFetcher<?>> configurer) {
+      this.mutationConfiguration = configurer;
+      return this;
+    }
+
+    public GraphQLEngineBuilder build() {
+      return new GraphQLEngineBuilder(this);
+    }
   }
```
henneberger, line 103: This pattern doesn't conform to the current architecture, let's choose one way to do data fetchers. The current pattern is the `context` produces the data fetchers. We could move to a typed GraphQLEngineBuilder for both servlet types and move it out of the context, but we shouldn't have a quasi pattern where we construct query fetchers one way and subscriptions/mutation fetchers a different way.

henneberger (changes requested):
Please make the graphql model backwards compatible

PR #851, opened 2024-10-02
Add udf support to run/test

henneberger:


sqrl-testing/sqrl-integration-tests/src/test/resources/snapshots/myudf/myudf.txt
```diff
@@ -0,0 +1,2 @@
+>>>MyTable.graphql
+{"data":{"MyTable":[{"val":1},{"val":2},{"val":3},{"val":4},{"val":5},{"val":6},{"val":7},{"val":8},{"val":9},{"val":10}]}}
```
mbroecheler, line 2: If MyScalarFunction adds the two arguments, shouldn't those values be double what they are?

mbroecheler: Ah, you are only returning "val". Can you add myFunc instead so we can test that the function actually gets invoked?

mbroecheler (approved):
One minor comment on the test

PR #852, opened 2024-10-03
Adding some more logging to the tests

Zesky665:
Added a log to the usecases tst runner, it will make reading the logs easier. 
Previously there was no easy way to check which usecase the output belonged to.


henneberger (approved):
Looks great!

PR #858, opened 2024-10-10
Specify additional sources when creating a package

nandorsoma:
When publishing a package, files are taken from the package root. However, additional sources, like build-generated files (e.g., .jar), may need inclusion. For instance, when publishing a UDF, metadata is stored in the root, while the UDF itself is compiled into a .jar in the target folder. Instead of using a separate script to assemble the package before publishing, it is more robust to include additional sources directly in the package.json. See an example https://github.com/DataSQRL/sqrl-functions/pull/1.

### DIFF
--- a/sqrl-base/src/main/java/com/datasqrl/config/PackageConfiguration.java
+++ b/sqrl-base/src/main/java/com/datasqrl/config/PackageConfiguration.java
@@ -29,6 +29,8 @@ public interface PackageConfiguration {
 
   List<String> getTopics();
 
+  List<String> getSources();
+
   void checkInitialized();
 
   Dependency asDependency();
--- a/sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/PackageConfigurationImpl.java
+++ b/sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/PackageConfigurationImpl.java
@@ -48,6 +48,8 @@ public class PackageConfigurationImpl implements PackageConfiguration {
   String description = "";
   @Default
   List<String> topics = List.of();
+  @Default
+  List<String> sources = List.of();
 
   public void checkInitialized() {
     Preconditions.checkArgument(!Strings.isNullOrEmpty(getName()) &&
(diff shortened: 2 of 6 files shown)


mbroecheler:
The problem you are trying to solve is having to copy-paste the jar file into the root directory before publishing, correct?
I wonder if we could solve this with an additional argument `--source` to the `publish` command that effectively does the same thing.
The downside of this solution is that you have to specify the source when you run the command.
The upside of this solution is that we don't have to include build specific information into the package manifest which should ideally only include information on what is in the package and not how the package was assembled.
WDYT @nandorsoma ?

nandorsoma:
> The problem you are trying to solve is having to copy-paste the jar file into the root directory before publishing, correct? I wonder if we could solve this with an additional argument `--source` to the `publish` command that effectively does the same thing. The downside of this solution is that you have to specify the source when you run the command. The upside of this solution is that we don't have to include build specific information into the package manifest which should ideally only include information on what is in the package and not how the package was assembled. WDYT @nandorsoma ?

Thanks for the review, @mbroecheler!

That’s definitely an option. I personally feel that listing the sources fits within the scope of “what’s in the package.” However, I don’t have a strong preference and would be open to that approach.

What concerns me more is that the current method combines the root and the additional sources from the manifest, which feels like a patchwork solution. What do you think about using --source but making it exclusive? So, only specified files/folders would be added, and if you want the root directory plus another file, you’d specify --source "." "./target/udf.jar". This could address the issue raised [here](https://github.com/DataSQRL/sqrl/issues/859).

mbroecheler:
How about a `-s` `--sources` option that is a list of paths which defaults to the current path (`.`). Then it's backwards compatible to how it currently works and you can make it work how you want by specifying different sources.

This seems the most user friendly approach since - in most cases - you just want to publish what is in the current directory. WDYT?

henneberger (changes requested):
See @mbroecheler comments in main conversation

mbroecheler:
removed publishing functionality.

PR #868, opened 2024-10-14
Support generic blackhole sink connector

henneberger:


sqrl-base/src/main/java/com/datasqrl/config/SystemBuiltInConnectors.java
```diff
@@ -13,7 +13,8 @@ public enum SystemBuiltInConnectors {
   PRINT_SINK(Name.system("print"), ExternalDataType.sink),
   LOCAL_FILE_SOURCE(Name.system("localfile"), ExternalDataType.source),
   LOGGER(Name.system("logger"), ExternalDataType.sink),
-  LOG_ENGINE(Name.system("log"), ExternalDataType.sink);
+  LOG_ENGINE(Name.system("log"), ExternalDataType.sink),
+  ;
```
nandorsoma, line 17: Theres an extra `,` here.

nandorsoma (approved):
I have one minor comment, but overall, it looks good to me. One more high lvl thing came to mind: what’s the benefit of having a blackhole sink? Flink will still compute the output, which adds overhead. In contrast, when there’s no sink for a table or it doesn’t have a GraphQL interface, the DAG planner prunes it. This is one of the advantages of SQRL. So, what’s the reasoning behind bypassing this behavior with a blackhole sink?

sqrl-base/src/main/java/com/datasqrl/config/SystemBuiltInConnectors.java
henneberger: Yeah, I usually add one so it helps reduce the number of diff lines

henneberger:
> I have one minor comment, but overall, it looks good to me. One more high lvl thing came to mind: what’s the benefit of having a blackhole sink? Flink will still compute the output, which adds overhead. In contrast, when there’s no sink for a table or it doesn’t have a GraphQL interface, the DAG planner prunes it. This is one of the advantages of SQRL. So, what’s the reasoning behind bypassing this behavior with a blackhole sink?

Good question. I use it primarily for testing. We also suggest people to replace their external sinks with blackhole sinks when the write local test cases. We could prune the job graph, but running with the test goal is usually additive to the job graph rather than subtractive. Wdyt?

PR #920, opened 2024-10-17
Resolve Test Error Logging and Repair Broken Integration Tests

nandorsoma:


Zesky665 (approved):

henneberger:
Break color formatting? Maybe we can log twice, once for the logger and once at the end for the user. Also why is the analytics case flakey?

sqrl-tools/sqrl-test/src/main/java/com/datasqrl/DatasqrlTest.java
```diff
@@ -174,29 +175,25 @@ public int run() {
...
           exitCode = 1;
         } else if (e instanceof JobFailureException) {
-          JobFailureException ex = (JobFailureException) e;
-          logRed("Flink job failed to start.");
-          ex.printStackTrace();
+          log.error("Flink job failed to start.", e);
           exitCode = 1;
         } else if (e instanceof MissingSnapshotException) {
           MissingSnapshotException ex = (MissingSnapshotException) e;
-          logRed("Snapshot on filesystem but not in result: " + ex.getTestName());
```
henneberger, line 188: This appears to break our color formatting. This is really helpful to see. We spoke about maybe logging twice so it can be easier to find in the build logs, where we could log when we observe a test failure and the produce a color summary at the end.

henneberger (changes requested):
See comments

henneberger:
What is this PR solving? Is there an issue attached to this?

PR #966, opened 2024-11-05
Avro improvements

### LINKED ISSUE #965: Use 'legacy timestamp mapping' by default for avro schema when using non-avro formats
For example, when using the json format, we should always apply the legacy timestamp mapping.

henneberger:
Fixes #965


mbroecheler (approved):

henneberger:
Broken up into smaller commits

PR #970, opened 2024-11-07
Revert avro schema to original schema mapping

henneberger:


henneberger:
We don't need to revert the whole schema mapping

PR #999, opened 2024-11-13
Allow skipping records that don't deserialize

henneberger:


mbroecheler:
We are addressing this in flink-sql-runner. Leave open for inspiration until this is fully fixed there.

mbroecheler:
Fixed in flink-sql-runner

PR #1019, opened 2024-12-03
Pre-loaded Profiles

mbroecheler:
Currently, we have to download the default profile from the remote repository in order to compile a SQRL project. This has a number of downsides:
1) We have to remember to always publish a profile for the current version as part of the release
2) If the user is offline it won't work

This draft shows how we could ship default profiles through a `StaticRepository` so they can be included in the release directly and don't require repository access.


mbroecheler:
Superseded by #1078

PR #1020, opened 2024-12-03
Flink aggregate functions tests

Zesky665:
The NTILE and ARRAYAGG functions returned errors suggesting that they are unsupported.


mbroecheler:
We should preserve some of this test in the DagPlannerTest

mbroecheler:
Since we use the flink parser, we no longer need to test it

PR #1022, opened 2024-12-04
Explicitly set nullability

mbroecheler:


sqrl-calcite/src/main/java/com/datasqrl/util/CalciteUtil.java
```diff
@@ -452,7 +452,7 @@ public static void addFilteredDeduplication(RelBuilder relB, int timestampIdx, L
 
     //Compute the lag for all columns other than the order and timestamp column, ordered by timestamp
     Function<Integer, RexNode> lagFunction = idx -> rexBuilder.makeOver(
```
henneberger, line 454: relBuilder has an 'aggregateCall' builder for making rex ops which has extra return type inference built in. In this specific issue, we're not looking at the lower bound frame when doing the inference.

see https://github.com/apache/calcite/blob/c49792f9c72159571f898c5fca1e26cba9870b07/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java#L6072

and

https://github.com/apache/calcite/blob/ec0dc3c886ef06020294dc80971c29ec3b90fa44/core/src/main/java/org/apache/calcite/tools/RelBuilder.java#L4649

mbroecheler (commented):
Looks good, ready for merge

PR #1024, opened 2024-12-05
Flink conditionals functions test

Zesky665:
The following functions don't seem to be supported:
- IS_ALPHA()
- IS_DECIMAL()
- IS_DIGIT()
- GREATEST(NULL)
- LEAST(NULL) 
- IFNULL()
- COALESCE(NULL, NULL)
- NULLIF - seems to work correctly, but NULL values are not accepted


mbroecheler:
Since moving to FlinkSQL as the parser, we no longer have to test compatibility at this level.

PR #1026, opened 2024-12-07
Flink type conversion functions test

Zesky665:
There are a lot of functions here that don't work as expected.


mbroecheler:
Since moving to FlinkSQL as the parser, we no longer have to test compatibility at this level.

PR #1028, opened 2024-12-09
Flink grouping functions test

Zesky665:
Tests for the grouping system functions.


mbroecheler:
This is a very nicely written test, we should preserve it in the Dagplannertest

PR #1029, opened 2024-12-11
Add DoubleToVector to the std library

nandorsoma:
DoubleToVector is missing from the vector module making it unable to import. This PR fixes that.


mbroecheler (approved):

henneberger:
Reverting because it is failing the build.

PR #1031, opened 2024-12-18
Simplify Transpiler

mbroecheler:
by doing some lightweight string manipulation for our custom statements and otherwise using Flink entirely for parsing and planning.

This has the potential to eliminate a lot of code (parser and Sqrl2Flink translation) while making it easier to remain compatible with Flink.

It also allows us to move to FLinkSQL for connector configuration to replace our table.json files.

Open questions:
- can we support SQRL's table function syntax somehow?

To Dos:
[] inject SqrlLogicalPlanRewriter into the ViewConverter and keep track of metadata.
[] support CREATE TABLE for mutations
[] make the lightweight statement parser more robust using SqlIdentifier and ideally some real parsing instead of regex


mbroecheler:
Superseded by #1078

PR #1034, opened 2025-01-02
Attempt to start a compilation server for compile command

velo:
I decided to spent some time getting compiler faster.
I went with Daniel suggestions of making an HTTP server to handle compile in the same JVM.... numbers are not that impressive.

I built that creditcard-rewards.sqrl example:

```
HTTP server started on port 7775
14:38:09.950 [vert.x-worker-thread-0] INFO  com.datasqrl.cmd.CompilationServerCommand - Compilation finished after 14s
14:38:47.578 [vert.x-worker-thread-0] INFO  com.datasqrl.cmd.CompilationServerCommand - Compilation finished after 7s
14:40:09.908 [vert.x-worker-thread-0] INFO  com.datasqrl.cmd.CompilationServerCommand - Compilation finished after 7s
14:40:16.705 [vert.x-worker-thread-0] INFO  com.datasqrl.cmd.CompilationServerCommand - Compilation finished after 6s
```

Second compilation is faster, but not that fast

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
     <log4j.version>2.23.1</log4j.version>
     <scala.version>2.12</scala.version>
     <guava.version>31.1-jre</guava.version>
-    <lombok.version>1.18.24</lombok.version>
+    <lombok.version>1.18.34</lombok.version>
     <commonsio.version>2.11.0</commonsio.version>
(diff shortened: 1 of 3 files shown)


henneberger:
Lets remove this from the cli and just have another main method that we can invoke. This shouldn't be a customer facing feature.

PR #1039, opened 2025-01-06
Remove custom test goal from graphql generation

henneberger:
There is snapshot impact because now custom datatypes (e.g. int/float mapping) not backwards compatible. Also custom mutations would need to be extracted since they don't appear in the sqrl script, and subscriptions are not automatically mapped either.


mbroecheler:
Now longer needed in 0.6

Issue #909, opened 2024-10-17
Numeric primary key columns should not be mapped to Float in GraphQL


henneberger:
because this causes issues on equality comparison which is common for primary keys.

Options:
1. Just map them onto Int for now and detect overflow
2. Introduce BigInt to GraphQL as scalar for this

echauchot:
I think I would prefer option 2 as it does not require to deal with maintaining the equality in Int overflow case.

@henneberger, can you assign me the ticket ?

echauchot:
Need to solve [this ticket](https://github.com/DataSQRL/sqrl/issues/744) as a prerequisite

echauchot:
> I think I would prefer option 2 as it does not require to deal with maintaining the equality in Int overflow case.
> 
> @henneberger, can you assign me the ticket ?

I'm now a member of the org, I self assigned

mbroecheler:
This has been merged into main on a separate branch

Issue #744, opened 2024-08-06
Add graphql-java-extended-scalars by default


henneberger:
This maps better to our types and is supported by various graphql engines. We should add an option to fall back to native graphql, which has some nuances like treating ints as floats.

echauchot:
@henneberger can you assign the ticket to me ?

echauchot:
> @henneberger can you assign the ticket to me ?

I'm now member of the org, I self assigned

Issue #895, opened 2024-10-17
Fix timestamps in test runner and add back to tests


henneberger:
This [PR](https://github.com/DataSQRL/sqrl/pull/838/files) removed a number of timestamp fields from various test cases to work around a current limitation in the test runner:

- set a fixed timezone in the test runner (UTC)
- re-add those fields to the various tests

echauchot:
Snapshots are generated locally and committed to serve as reference snapshots for comparison in tests. The problem is that these snapshots use the local timezone that might be different from the one of the CI.  It leeds to comparison failures with the actual snapshots that use the CI timezone. Setting a fixed UTC timezone should fix this issue.

mbroecheler:
This has been fixed.

PR #1041, opened 2025-01-17
Support BigInteger in GraphQL (#744 and #909)

echauchot:
This PR covers both #744 and #909. Here is what it does:

- Registers extended graphQL scalars /BigInteger in the server (always, no conf item): reused and renamed unused addlTypes to extendedScalarTYpes
- Add a new **extendedScalarTypes** configuration in package.json (in the compiler configuration) which defaults to false for backward compatibility. So the default package.json does not change
- Registers extended graphQL scalars /BigInteger in the compiler based on the package.json conf item
- Reference new extendedScalarTypes in the specs doc
- map all Calcite BIGINT types to ExtendedScalars.GraphQLBigInteger based on the compiler config parameter
- Add a new usecase test with BigInteger support enabled: **simple-with-bigint-support** that does a simple select on the ecommerce/Orders  table with BIGINT as primary key
- Tested with both graphQL schema provided and not provided on UseCaseCompileTest and committed the corresponding snapshot 
- Tested with both graphQL schema provided and not provided with test and run goals on FullUsecasesIT and committed the corresponding snapshot 

@henneberger @mbroecheler PTAL


henneberger (approved):
Looks good. One comment about naming.

echauchot:
> Looks good. One comment about naming.

@henneberger thanks for your review. I don't see your naming comment, have you published your review ?

echauchot:
What is strange regarding the build errors is that **UseCaseCompileTest** runs fine when run alone but part of mvn verify, some tests start to fail:
- **simple-with-bigint-support**: fails with a snapshot mistmatch. It is expecting a  graphql schema with BigInteger but the generated schema still uses Float as if the use case was using the default package.json (with extendedScalarTypes disabled) instead of the packaged one.
- **seedshop-avro**: fails with _Not a valid SQRL data type_ in TableConverter.sourceToTable() with both provided package.json

sqrl-calcite/src/main/java/com/datasqrl/calcite/QueryPlanner.java
```diff
@@ -203,9 +203,15 @@ public RelDataType parseDatatype(String datatype) {
       return this.cluster.getTypeFactory().createSqlType(SqlTypeName.TIMESTAMP, 3);
     }
 
+    if (datatype.equalsIgnoreCase("BigInteger")) {
+      return this.cluster.getTypeFactory().createSqlType(SqlTypeName.BIGINT);
```
henneberger, line 207: We could use the graphql-java base set of scalars (`scalar GraphQLBigInteger`) so all of the types can be mapped uniformly. https://github.com/graphql-java/graphql-java-extended-scalars?tab=readme-ov-file#java-primitives

echauchot: What do you mean ?
`scalar GraphQLBigInteger` maps to `"BigInteger"`  String litteral in [JavaPrimitives](https://github.com/graphql-java/graphql-java-extended-scalars/blob/9e283115657220fd191e7a03d9ecf07964746aed/src/main/java/graphql/scalars/java/JavaPrimitives.java#L373)

echauchot: I pushed comments and test related to that

echauchot:
> What is strange regarding the build errors is that **UseCaseCompileTest** runs fine when run alone but part of mvn verify, some tests start to fail:
> 
> * **simple-with-bigint-support**: fails with a snapshot mistmatch. It is expecting a  graphql schema with BigInteger but the generated schema still uses Float as if the use case was using the default package.json (with extendedScalarTypes disabled) instead of the packaged one.
> * **seedshop-avro**: fails with _Not a valid SQRL data type_ in TableConverter.sourceToTable() with both provided package.json

My bad, my local m2 repo was somehow out of sync. With a full maven build, there is no error except indeed the timestamp mismatch in FullUseCasesIT: the new use case generated a snapshot with my local timezone which is different from the timezone the git servers are in.

We should make the tests timezone agnostic so that the locally generated snapshots are comparable to the expected ones in the test infrastructure. I suggest programmatically forcing the timezone in the tests for both snapshot generation and comparison. WDYT ?

In the meantime, I'll update the snapshot for the new test to the expected one.

echauchot:
> > What is strange regarding the build errors is that **UseCaseCompileTest** runs fine when run alone but part of mvn verify, some tests start to fail:
> > 
> > * **simple-with-bigint-support**: fails with a snapshot mistmatch. It is expecting a  graphql schema with BigInteger but the generated schema still uses Float as if the use case was using the default package.json (with extendedScalarTypes disabled) instead of the packaged one.
> > * **seedshop-avro**: fails with _Not a valid SQRL data type_ in TableConverter.sourceToTable() with both provided package.json
> 
> My bad, my local m2 repo was somehow out of sync. With a full maven build, there is no error except indeed the timestamp mismatch in FullUseCasesIT: the new use case generated a snapshot with my local timezone which is different from the timezone the git servers are in.
> 
> We should make the tests timezone agnostic so that the locally generated snapshots are comparable to the expected ones in the test infrastructure. I suggest programmatically forcing the timezone in the tests for both snapshot generation and comparison. So I updated the test to do that.
> 
> Now the timestamps in all the snapshots are exactly the ones that are inserted in the synthetic source. Before that there was a difference that might have been a bug. So I updated all the snapshots for all the use cases with the timezone forcing.

As discussed, I'm no more updating all the snapshots and trying to fix the timezone issue as it is part of a broader timezone plan.  I replaced BigIntOrdersTest.snapshot with the snapshot generated in the CI (with UTC-8 timezone) to workaround this local timezone issue.

Regarding a ticket opening for fixing timezone issue, I found [#895](https://github.com/DataSQRL/sqrl/issues/895) which mentions setting a fixed timezone to UTC in the tests. This tickets seems a good fit to me so I just added a comment to it to trace the observed behavior in this PR.

sqrl-calcite/src/main/java/com/datasqrl/calcite/QueryPlanner.java
henneberger: Calling it BigInteger may be confusing since there is also a "BigInt" that is a java Long type. The Numeric type is the type that maps to java BigInteger. Since we have one type system for both sql and graphql, it may be better to disambiguate this by calling it "GraphQLBigInteger"

echauchot: Thanks it is clearer. I agree. I pushed a fix. PTAL

echauchot:
signed all the commits and pushed

echauchot:
> signed all the commits and pushed

But when we hit github button rebase and merge, the signatures seem to be gone :smile:

henneberger (approved):

echauchot:
@henneberger thanks for your review

PR #1044, opened 2025-01-23
Document server-core server-vertx

echauchot:
- some refactorings or duplicated code and some cleaning
- added some javadocs to major classes of sqrl-server-core and sqrl-server-vertx modules
- started to add a counter for pages and non-pagedjdbc queries.

This PR is not ready for review, just used as a base for discussion if needed.

@mbroecheler @henneberger PTAL


echauchot:
@henneberger I removed the counter to use that PR as doc only. PTAL

PR #1047, opened 2025-02-03
Remove profiles support

velo:
Initial PR just removing profiles support


sqrl-tools/sqrl-cli/src/main/java/com/datasqrl/cmd/AbstractCompilerCommand.java
```diff
@@ -62,21 +62,21 @@ public abstract class AbstractCompilerCommand extends AbstractCommand {
 
   @CommandLine.Option(names = {"--profile"},
       description = "An alternative set of configuration values which override the default package.json")
-  protected String[] profiles = new String[0];
+  protected String[] _profiles = new String[0];
```
mbroecheler, line 65: We can remove the setting from the compiler for 0.6 since we can make breaking changes

sqrl-tools/sqrl-cli/src/test/java/com/datasqrl/cmd/AssertStatusHook.java
```diff
@@ -10,6 +10,7 @@ public class AssertStatusHook implements StatusHook {
   private boolean failed;
   @Getter
   private String failMessage = null;
+  private Throwable failure;
```
mbroecheler, line 13: Did you add that for troubleshooting?

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfigCommons.java
```diff
@@ -419,6 +419,12 @@ public static SqrlConfig fromFiles(ErrorCollector errors, String jsonSchemaResou
         throw localErrors.handle(e);
       }
     }
+
+    try {
```
mbroecheler, line 423: If you do the default configuration last, it will overwrite the configuration the user provides, right? We want to make sure the user configuration takes precedence.

In general, this is not the right method to hard code the default configuration into. Maybe you can relax the method to accept a List<URL> as argument and then make adding the default configuration happen one level higher (i.e. in the caller). 
In the future, we may have to add a configuration option to ignore the default configuration if the user wants to use just their own (and doesn't want to have to overwrite everything).

sqrl-tools/sqrl-packager/src/main/java/com/datasqrl/cmd/PackageBootstrap.java
```diff
@@ -45,81 +47,24 @@ public PackageJson bootstrap(Path rootDir, List<Path> packageFiles,
...
     existingPackage.ifPresent(configFiles::addAll);
 
     // Could not find any package json
+    PackageJson packageJson;
     if (configFiles.isEmpty()) {
-      throw new RuntimeException("Could not find package.json");
+        packageJson = createDefaultConfig(errors);
+    } else {
+        // Merge all configurations
+        packageJson =  SqrlConfigCommons.fromFilesPackageJson(errors, configFiles);
```
mbroecheler, line 65: I wonder if we can create the list of package.json in this method and simply append the default package json to list. Then you wouldn't have to have a special case for createDefaultConfig.

mbroecheler (changes requested):
Good work. Some comments inline

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfigCommons.java
mbroecheler: Can you add a test case for merging multiple configurations. I'm not sure we have test coverage for that right now.

mbroecheler (changes requested):
this is mostly there - thank you very much. See comments for handling of the default package json and merging of use provided package.json files.

sqrl-tools/sqrl-cli/src/test/java/com/datasqrl/cmd/AssertStatusHook.java
velo: Yes, I had a few tests failing and the original exception was not obvious, so I added this so I would fetch it later

sqrl-tools/sqrl-packager/src/main/java/com/datasqrl/cmd/PackageBootstrap.java
velo: since "default-package.json" is backed into jar, accessing it is not that straight forward (it should be, cause Paths can refer entries inside zip files), but when I tried to convert the resource url into Path it worked sometimes, but didn't other times leading to unstable tests, that is why I'm using the fromURL method.

But I can change "fromFilesPackageJson" to allow empty, which would effectively be the same

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfigCommons.java
velo: New test created, just checked if overriding enabled-engines works, and it does work

sqrl-testing/sqrl-integration-tests/src/test/java/com/datasqrl/discovery/FlexibleSchemaInferencePreprocessorTest.java
```diff
@@ -41,8 +42,7 @@ public class FlexibleSchemaInferencePreprocessorTest extends AbstractAssetSnapsh
   protected FlexibleSchemaInferencePreprocessorTest() {
     super(FILES_DIR.resolve("output"));
     try {
-      packageJson = SqrlConfigCommons.fromFilesPackageJson(errors,
-          List.of(Path.of("../../profiles/default/package.json")));
+      packageJson = new PackageJsonImpl(SqrlConfigCommons.fromURL(errors, getClass().getResource("/default-package.json")));
```
mbroecheler, line 45: replace with SqrlConfigCommons.getDefaultPackageConfig()
which in turn calls SqrlConfigCommons.getPacakgeConfig(errors, List.of())

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfigCommons.java
```diff
@@ -402,12 +402,12 @@ public static SqrlConfig fromFiles(ErrorCollector errors, @NonNull Path firstFil
   }
 
   public static SqrlConfig fromFiles(ErrorCollector errors, String jsonSchemaResource, @NonNull List<Path> files) {
```
mbroecheler, line 404: rename to getPackageConfig?

sqrl-tools/sqrl-packager/src/main/java/com/datasqrl/cmd/PackageBootstrap.java
```diff
@@ -156,16 +95,7 @@ public PackageJson bootstrap(Path rootDir, List<Path> packageFiles,
...
-
-    return setDefaultConfig(packageJson);
-  }
-
-  public PackageJson setDefaultConfig(PackageJson packageJson) {
-    packageJson.setProfiles(new String[]{"datasqrl.profile.default"});
-    packageJson.getDependencies()
-        .addDependency("datasqrl.profile.default",
-            new DependencyImpl("datasqrl.profile.default", "0.5.10", "dev"));
+    PackageJson packageJson = new PackageJsonImpl(SqrlConfigCommons.fromURL(errors, getClass().getResource("/default-package.json")));
```
mbroecheler, line 98: replace with SqrlConfigCommons.getDefaultPackageConfig()

mbroecheler (approved):
Looks good. I have one minor suggestion for how to make the methods more obvious and have less code duplication.

sqrl-tools/sqrl-packager/src/main/java/com/datasqrl/cmd/PackageBootstrap.java
velo: Actually, this public method was not been used, I removed it.

velo:
```
$ docker run -ti --rm  ghcr.io/datasqrl/cmd:pr-1047 compile --help
Compiling...this takes about 10 seconds
Unknown option: '--help'
Usage: datasqrl compile [-t=<targetDir>] [-a=<generateAPI>]...
                        [-c=<packageFiles>]... [<files>...]
Compiles an SQRL script and produces all build artifacts
      [<files>...]           Main script and (optional) API specification
  -a, --api=<generateAPI>    Generates the API specification for the given type
  -c, --config=<packageFiles>
                             Package configuration file(s)
  -t, --target=<targetDir>   Target directory for deployment artifacts
```

Just tested my latest snapshot, working as expected. `--profiles` is gone

PR #1051, opened 2025-02-06
Check how sqrl behaves on java 21

velo:


velo:
Flink 1.16 is only compatible with java 11, so this is not possible.

Flink 1.17 onwards are only compatible with java 17

PR #1076, opened 2025-03-30
Fix/rowtime nullability

mbroecheler:


velo:
Going to merge as is, as this fixes major issue with build, going to create a follow up PR to fix the sqrl test issue

PR #1082, opened 2025-04-11
bugfixes

mbroecheler:
1) nullability of GraphQL parameters
2) processing of import hints


sqrl-planner/src/main/java/com/datasqrl/v2/parser/SqrlStatementParser.java
```diff
@@ -117,15 +117,15 @@ public SQLStatement parseStatement(String statement) {
...
+        return new SqrlImportStatement(packageIdentifier, alias, comment);
       } else if (directive.equalsIgnoreCase("export")) {
         Matcher subMatcher = EXPORT_PARSER.matcher(body);
         checkFatal(subMatcher.find(), ErrorCode.INVALID_IMPORT, "Could not parse IMPORT statement");
         ParsedObject<NamePath> packageIdentifier = parseNamePath(subMatcher, "package", statement);
         ParsedObject<NamePath> tableName = parseNamePath(subMatcher, "identifier", statement);
         checkFatal(packageIdentifier.isPresent(), ErrorCode.INVALID_EXPORT, "Missing package path");
         checkFatal(packageIdentifier.isPresent(), ErrorCode.INVALID_EXPORT, "Missing table");
-        return new SqrlExportStatement(tableName, packageIdentifier);
+        return new SqrlExportStatement(tableName, packageIdentifier, comment);
```
velo, line 128: Do we need a new test covering export hints?

I look for `/*+` and the only change I found is on the imports.

velo (approved):
Overall looks good, just raised one question about export testing

sqrl-planner/src/main/java/com/datasqrl/v2/parser/SqrlStatementParser.java
mbroecheler: Great observation - thank you for being so diligent.
We currently do not have any hints that apply to exports. I just implemented the infrastructure to support that (since I was working on it), but there aren't any hints yet that we could test it with.

PR #1084, opened 2025-04-16
fix validation of parameter nullability, fix validation messages.

mbroecheler:


velo (approved):
LGTM, the only thing I'm missing would be a graphql test that would read data with and without the optional field.

PR #1085, opened 2025-04-18
Simplify queries in Server & fix query lookup

mbroecheler:
This PR greatly simplifies how we handle Sql query execution on the server, uses prepared queries and variable binding for limit and offset, and allow for null parameters.


sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/jdbc/DatabaseType.java
```diff
@@ -0,0 +1,20 @@
+package com.datasqrl.graphql.jdbc;
+
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public enum DatabaseType {
```
velo, line 6: What if we move this to a common/model project 

That way sqrl-base don't need to import vertex module and all it's dependencies

mbroecheler: Good point. This is temporary.
I want to get rid of sqrl-base when we do the spring cleaning. It doesn't serve a purpose anymore. 

- We need the server-core module to be very lightweight since that runs in production.
- We need the flink libraries (which are mostly in the runner now) to be very lightweight since that runs in production.
- The rest is for the compiler which is a beast anyway (includes all of Flink, Calcite, etc) so we can consolidate everything in there.

In other words, there is no need for a shared "base" library anymore. We just have Flink specific modules. Vert.x specific modules. And the rest should be consolidated into planner and then we have the packager modules for instrumentation, CLI, etc.
We will plan this out next week.

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/jdbc/AbstractQueryExecutionContext.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  public List getParamArguments(List<QueryParameterHandler> parameters) {
+    List paramObj = new ArrayList(parameters.size()+2);
+    for (int i = 0; i < parameters.size(); i++) {
+      QueryParameterHandler param = parameters.get(i);
+      Object o = param.accept(this, this);
+      paramObj.add(o);
+    }
+    assert paramObj.size() == parameters.size();
```
velo, line 47: I though I was the only person that would use assertions to leave breadcrumbs behind, would be nice to include a message =)

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/jdbc/AbstractQueryExecutionContext.java
```diff
@@ -0,0 +1,55 @@
...
+  public Object visitArgumentParameter(ArgumentParameter argumentParameter,
+      QueryExecutionContext context) {
+    return context.getArguments().stream()
+        .filter(arg -> arg.getPath().equalsIgnoreCase(argumentParameter.getPath()))
+        .findFirst()
+        .map(Argument::getValue)
+        .orElse(null);
+  }
+
+  public List getParamArguments(List<QueryParameterHandler> parameters) {
```
velo, line 40: you could do this using streaming api =)

velo (approved):
Nagged in a couple of places, nothing critical

PR #1087, opened 2025-04-22
Migrate code to java 17

velo:


sqrl-base/src/main/java/com/datasqrl/config/ConnectorConf.java
```diff
@@ -35,10 +36,18 @@ class Context {
 
     public Map<String,String> toVariables() {
       Map<String,String> vars = new HashMap<>(variables);
-      if (tableName!=null) vars.put("table-name", tableName);
-      if (origTableName!=null) vars.put("original-table-name", origTableName);
-      if (filename!=null) vars.put("filename", filename);
-      if (format!=null) vars.put("format", format);
+      if (tableName!=null) {
```
mbroecheler, line 39: You might have to run a code formatter on this @velo

PR #1108, opened 2025-05-05
Consolidate modules

velo:
- Consolidate modules and simplify module structure
- Consolidate sqrl-base, sqrl-calcite, sqrl-errors, sqrl-name into sqrl-planner
- Remove dead code


velo:
+ 14% code coverage, not bad
image

mbroecheler (approved):
Thanks for moving all this around. Hard to do a careful code review, but if the tests pass this is fine.
Did some investigating of what might be causing the change in function identifier escaping and as best I can tell, this has to do with a change in Calcite which is no uniformly quoting all identifiers - functions included. I think this is probably for the better since it makes the behavior uniform.

PR #1255, opened 2025-05-23
Import acorn-graphql into sqrl

velo:


velo:
@mbroecheler I see you already based some work (MCP) on this, am I free to merge it?


Cheers

PR #1290, opened 2025-06-13
Update kafka topic name to take template into account

velo:


sqrl-planner/src/main/java/com/datasqrl/planner/tables/FlinkTableBuilder.java
```diff
@@ -203,4 +203,8 @@ public static FlinkTableBuilder toBuilder(SqlCreateTable table) {
         table.getPartitionKeyList(),
         table.getWatermark().orElse(null));
   }
+
+  public String getTopicName() {
+    return getConnectorOptions().get("topic");
```
ferenc-csaky, line 208: I was wondering if we can make this smarter, to not create a map for one topic, but at this point it would end up some code duplication, cause iterating over `SqlNodes` are necessary anyways, and creating a map that is dereferenced right away can pretty much be ignored, so I came to agreement that this is fine. :)

mbroecheler (approved):

PR #1300, opened 2025-06-17
Adapt functions and types to `flink-sql-runner` refactor

ferenc-csaky:
The `functions` and `types` modules are now merged under an `stdlib` module in `flink-sql-runner`, this change adapts the logic in the repo. This also applies to the sub-modules and the package structure.

### Relevant points:
- Refactored maven modules
- Refactored function and type packages
- Adapted tests

### TODO
- [ ] Remove the `sqrl-flink-runner` docker image


ferenc-csaky:
@mbroecheler @velo I think we can get rid of the `sqrl-flink-runner` img completely. The only thing it adds to the `flink-sql-runner` img is `COPY sqrl-functions/*.jar /opt/flink/lib/`, but `sqrl-functions` does not exist anymore.

Although right now I do not copy `stdlib-commons` to the flink `lib/` in the `flink-sql-runner` img, as I did not see it as a necessity.

WDYT?

mbroecheler:
I'll let @velo chime in as he build that image. From my perspective it would be nice if sql flink runner was completely self contained and works with the outputs from DataSQRL without requiring a separate image. That would make the two projects nicely complementary.

mbroecheler (approved):
Nicely done!!!

velo:
Hey, I'm all in for consuming a resource that is ready for me and zero effort on my part =)

I think we should only have the integration test running, just as a final check that all is working

PR #1311, opened 2025-06-19
Replace commons-configuration with jackson

velo:


sqrl-tools/sqrl-config/pom.xml
```diff
@@ -33,12 +33,12 @@
     </dependency>
     <!-- Configuration -->
     <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-configuration2</artifactId>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-annotations</artifactId>
```
ferenc-csaky: `jackson-databind` depends on `jackson-annotations`, so it can be removed from our POM.

sqrl-testing/sqrl-integration-tests/src/test/java/com/datasqrl/FullUseCasesIT.java
```diff
@@ -282,7 +282,8 @@ void useCase(UseCaseTestParameter param) {
             case SUCCESS_WITH_CONTENT:
               break;
             default:
-              fail("Flink job failed with: " + result.getResultKind());
+              org.assertj.core.api.Assertions.fail(
```
ferenc-csaky: nit: I do not see the relevance of this change, `org.assertj.core.api.Assertions.fail` is already static imported. If we change it, I'd rather remove the qualification @ L297

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SerializedSqrlConfig.java
ferenc-csaky, line 1: Since we modify this interface anyways, `public` keyword for the method definition is redundant and can be removed.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfigCommons.java
ferenc-csaky, line 1: In this new setup, I do not really see why we kept this class. Its delegating is hardcoded, in this form it won't be helpful fro another impl. IMO we should just call `SqrlConfig` everywhere, and remove `SqrlConfigCommons`.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
```diff
@@ -15,118 +15,575 @@
...
 
-/**
- * Interface for accessing configuration files that provides convenience methods for accessing
- * different data types and methods to handle errors in a way that preserves the error locality so
- * it is easy for users to understand where the error comes from and how to fix it.
- */
-interface SqrlConfig {
+  /** Create a new empty configuration with given version. */
+  public static SqrlConfig create(ErrorCollector errors, int version) {
+    ObjectNode base = MAPPER.createObjectNode();
```
ferenc-csaky: nit: `base` -> `root` to match the name with the class field.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
```diff
@@ -15,118 +15,575 @@
...
+  }
+
+  /** Load multiple JSON files and default-package.json, validate and merge them. */
+  public static SqrlConfig getPackageConfig(
+      ErrorCollector errors, String jsonSchemaResource, List<Path> files) {
+    boolean valid = true;
+    List<JsonNode> jsons = new ArrayList<>();
+
+    try {
+      URL url = SqrlConfigCommons.class.getResource("/default-package.json");
```
ferenc-csaky: Use `SqrlConfig.class` instead of `SqrlConfigCommons.class`.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
```diff
@@ -15,118 +15,575 @@
...
+      if (url != null) {
+        jsons.add(MAPPER.readTree(url));
+      }
+    } catch (IOException e) {
+      throw errors.exception("Error loading default configuration: %s", e.toString());
```
ferenc-csaky: Keeping this as `throw errors.withConfig("Error loading default configuration").handle(e);` seems more appropriate to me.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
```diff
@@ -15,118 +15,575 @@
...
+
+  /** Load multiple JSON files and default-package.json, validate and merge them. */
+  public static SqrlConfig getPackageConfig(
+      ErrorCollector errors, String jsonSchemaResource, List<Path> files) {
+    boolean valid = true;
+    List<JsonNode> jsons = new ArrayList<>();
+
+    try {
+      URL url = SqrlConfigCommons.class.getResource("/default-package.json");
+      if (url != null) {
```
ferenc-csaky: Is it okay to keep going if `default-package.json` is missing? Wondering if an exception should be thrown here instead.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
```diff
@@ -15,118 +15,575 @@
...
+        jsons.add(MAPPER.readTree(url));
+      }
+    } catch (IOException e) {
+      throw errors.exception("Error loading default configuration: %s", e.toString());
+    }
+
+    for (int i = 0; i < files.size(); i++) {
```
ferenc-csaky: Previously iteration was going backwards, is it okay to change that to iterate forward instead? If yes, this should be replaced with a for-each.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
```diff
@@ -15,118 +15,575 @@
...
+    JsonNode node;
+    try {
+      node = MAPPER.readTree(url);
+    } catch (IOException e) {
+      throw errors
+          .withConfig(url.toString())
+          .exception("Could not read JSON from URL [%s]: %s", url, e.toString());
+    }
+    ObjectNode merged = MAPPER.createObjectNode();
+    merge(merged, node);
```
ferenc-csaky: I do not see why is necessary. Here we pretty much copy over the same content from `node` to `merged`, since there is only 1 `node`. I believe this can be removed from both this method and `fromString`.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
```diff
@@ -15,118 +15,575 @@
...
+        jsons.add(MAPPER.readTree(file.toFile()));
+      } catch (IOException e) {
+        throw local.exception("Could not parse JSON file [%s]: %s", file, e.toString());
+      }
+    }
+    if (!valid) {
+      throw errors.exception("Configuration file invalid: %s", files);
+    }
+    ObjectNode merged = MAPPER.createObjectNode();
+    jsons.forEach(node -> merge(merged, node));
```
ferenc-csaky, line 170: #IMO the whole custom `merge` mehod is unnecessary, cause here we only append `ObjectNodes` anyways, we can simply use `setAll()`:
```java
ObjectNode merged = MAPPER.createObjectNode();
jsons.forEach(node -> merged.setAll((ObjectNode) node));
```

See my other comment regarding `fromURL` and `fromString`.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
```diff
@@ -15,118 +15,575 @@
...
+    List<String> keys = new ArrayList<>();
+    n.fieldNames().forEachRemaining(keys::add);
+    return keys;
+  }
 
-  static final int CURRENT_VERSION = 1;
-  static final String VERSION_KEY = "version";
+  public boolean containsKey(String key) {
+    JsonNode n = node();
+    return n != null && n.has(key) && !n.get(key).isContainerNode();
```
ferenc-csaky, line 277: Maybe it's intentional, but previously this method did not check for node type, so this may produce different behavior than before.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
```diff
@@ -15,118 +15,575 @@
...
     return as(key, Integer.class);
   }
 
-  default Value<Boolean> asBool(String key) {
+  public Value<Boolean> asBool(String key) {
     return as(key, Boolean.class);
   }
 
-  boolean hasKey(String key);
-
```
ferenc-csaky, line 116: The `Value` interface should be `public`, cause it is exposed publicly via other `public` methods.

ferenc-csaky (changes requested):
Added a couple suggestions/questions.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
velo: https://github.com/DataSQRL/sqrl/blob/configuration/sqrl-planner/src/main/java/com/datasqrl/error/ErrorCollector.java#L137

with config seems to expect a file, but that is not the case here

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
velo: `setAll` won't make the correct recursive merging.  it will just merge the first level but override everything 2 levels deep.

sqrl-tools/sqrl-config/src/main/java/com/datasqrl/config/SqrlConfig.java
velo: Added exception when url is null

PR #1326, opened 2025-06-25
Store Flink config in `FlinkPhysicalPlan`, generate YAML of its content

ferenc-csaky:
## Main changes
- Store Flink `Configuration` obj in `FlinkPhysicalPlan`, allow setting it via builder
- Generate a `DeploymentArtifact` from it, YAML string conversion via Flink util
- Base Flink config can be put into `default-package.json`
- Minor simplifications in `Sqrl2FlinkSQLTranslator`

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/planner/Sqrl2FlinkSQLTranslator.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/planner/Sqrl2FlinkSQLTranslator.java
@@ -95,7 +95,10 @@ import org.apache.calcite.sql.SqlOrderBy;
 import org.apache.calcite.sql.SqlSyntax;
 import org.apache.calcite.sql.parser.SqlParserPos;
 import org.apache.calcite.sql.validate.SqlNameMatchers;
+import org.apache.flink.api.common.RuntimeExecutionMode;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.ExecutionOptions;
+import org.apache.flink.configuration.PipelineOptions;
 import org.apache.flink.sql.parser.ddl.SqlAlterViewAs;
 import org.apache.flink.sql.parser.ddl.SqlCreateTable;
 import org.apache.flink.sql.parser.ddl.SqlCreateTableLike;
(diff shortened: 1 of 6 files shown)


sqrl-planner/src/main/java/com/datasqrl/planner/Sqrl2FlinkSQLTranslator.java
```diff
@@ -181,28 +184,28 @@ public Sqrl2FlinkSQLTranslator(
...
 
-    if (executionMode == ExecutionMode.STREAMING) {
-      sEnv.setRuntimeMode(org.apache.flink.api.common.RuntimeExecutionMode.STREAMING);
-      settingsBuilder.inStreamingMode();
-    } else {
-      sEnv.setRuntimeMode(org.apache.flink.api.common.RuntimeExecutionMode.BATCH);
-      settingsBuilder.inBatchMode();
+    // Init Flink config
+    var config = Configuration.fromMap(flink.getBaseConfiguration());
+    if (!jarUrls.isEmpty()) {
```
mbroecheler, line 190: Why is this only applied if there are jar URLs?

sqrl-planner/src/main/java/com/datasqrl/planner/Sqrl2FlinkSQLTranslator.java
```diff
@@ -181,28 +184,28 @@ public Sqrl2FlinkSQLTranslator(
...
+      config.set(
+          ExecutionOptions.RUNTIME_MODE,
+          executionMode == ExecutionMode.STREAMING
+              ? RuntimeExecutionMode.STREAMING
+              : RuntimeExecutionMode.BATCH);
     }
-    var tEnvConfig = settingsBuilder.build();
-
+    config.set(
```
mbroecheler, line 197: should this be in the above if clause instead?

sqrl-tools/sqrl-config/src/main/resources/default-package.json
```diff
@@ -3,6 +3,7 @@
   "enabled-engines": ["vertx", "postgres", "kafka", "flink"],
   "engines" : {
     "flink" : {
+      "config": {},
```
mbroecheler, line 6: I wonder what default configurations we want to put in here. Here is the list we currently have in DataSqrlRun:
```
    config.putIfAbsent("taskmanager.memory.network.max", "800m");
    config.putIfAbsent("state.backend.type", "rocksdb");
    config.putIfAbsent("table.exec.resource.default-parallelism", "1");
    config.putIfAbsent("rest.address", "localhost");
    config.putIfAbsent("rest.port", "8081");
    config.putIfAbsent("execution.target", "local"); // mini cluster
    config.putIfAbsent("execution.attached", "true"); // mini cluster
    if (isStreaming) {
      config.putIfAbsent("table.exec.source.idle-timeout", "1 s");
      config.putIfAbsent("execution.checkpointing.interval", "30 s");
      config.putIfAbsent("execution.checkpointing.min-pause", "20 s");
    }
```
I would love to move away from hardcoding these things in the runner and instead have them be:
1. In the default-package.json if they are used across most use cases
   * e.g. state.backend.type = rocksdb
2. Dynamically determined by the planner
   * e.g. idle-timeout when there are temporal joins (should default to 10 sec)
   * checkpoint interval and min-pause when the compilation goal is "test"
   * the other runner specific settings when the compilation goal is "run" or "test"
Another option would be to put the runner specific settings (last bullet point) and hard-code it into DataSqrlRun like we currently do.

WDYT?

sqrl-planner/src/main/java/com/datasqrl/planner/Sqrl2FlinkSQLTranslator.java
ferenc-csaky: Otherwise it will put an empty list into the YAML (`classpaths: []`), which I felt unnecessary and might also be misleading.

sqrl-planner/src/main/java/com/datasqrl/planner/Sqrl2FlinkSQLTranslator.java
ferenc-csaky: Yep, I reorganized the code and it's wrong now, so the idea is to put this into the `if` and keep the exec mode out. :)

sqrl-tools/sqrl-config/src/main/resources/default-package.json
ferenc-csaky: I agree with the suggested approach, and personally I think it's more simpler if we centralize the configuration setting as much as we can, so we do not end up magical end-results cause 3 different module touches the config at 5 different places.

PR #1329, opened 2025-06-26
Rework CLI and Flink configuration

ferenc-csaky:
## Main chages
- Merged `sql-run` into `sql-cli`
- Shaved ~200MB from base Docker img via reducing layers
- Shipping only 1 `sqrl-cli.jar` that conatins everything, adapted entrypoint script to that too
- Now we call everything form Java, the CLI is the entrypoint for every command
- Added Flink config to `FlinkPhysicalPlan`, and writing it out during compile
- In run/test commands, the YAML file is read back

## TODO
- [x] Check on the failing test cases


sqrl-planner/src/main/java/com/datasqrl/planner/Sqrl2FlinkSQLTranslator.java
```diff
@@ -181,28 +184,34 @@ public Sqrl2FlinkSQLTranslator(
...
+          jarUrls.stream().map(URL::toString).collect(Collectors.toList()));
+    }
 
     if (executionMode == ExecutionMode.STREAMING) {
-      sEnv.setRuntimeMode(org.apache.flink.api.common.RuntimeExecutionMode.STREAMING);
-      settingsBuilder.inStreamingMode();
-    } else {
-      sEnv.setRuntimeMode(org.apache.flink.api.common.RuntimeExecutionMode.BATCH);
-      settingsBuilder.inBatchMode();
+      config.addAll(flink.getStreamingSpecificConfig());
```
mbroecheler: We need to be careful to not overwrite user provided configuration with configuration we "infer".
The way we merge configuration should be:
1. Start with the "inferred" configuration
2. Overwrite with user provided configuration
3. Overwrite with configuration that takes precedence like the ExecutionMode

ferenc-csaky: Nice comment, you pinpointed the part which I was not sure about. So in the current `usecases` scripts there are examples that contain the Flink config under the `values.flink-config` key in the `package.json`, those are legacy?

My though would be that any `package.json` should have the schema of the current `default-package.json`, and Flink configuration should be placed under `engines.flink.config`, right?

And if that's correct my idea would be to do the following:
1. Initialize with the relevant content of `default-package.json`
2. Add anything that is defined by the user in `package.json`, that will overwrite anything that already exists.
3. Add parts that set by the SQRL platform anyways (like `ExecutionMode`, or temporal join related improvements, etc.)

I think that pretty much aligns with what you described, just tried to bring it closer to the actual impl. :)

WDYT?

ferenc-csaky: Now that I checked on it, `SqrlConfig` already works this way, it first reads `deafult-package.json`, then `package.json`, so any default values will already be overwritten if it is set by the user as is.

mbroecheler: Correct, the merging of default-package with user provided configs is done at the SqrlConfig layer - you don't need to do that.
What we need to be careful about is this:
If we are making default settings based on best practices or what we inferred from the script, those config options need to be set BEFORE we apply the user configuration.
For example, we set a default idle-timeout if we detect temporal joins. But if the user has configured it, we would currently overwrite the user configuration which is undesirable. Instead, we want to apply these "inferred" configurations first, then load the user configuration on top.
And finally, we add any user configuration that takes precedence over the "config" section. Currently, that's only the "mode" configuration.

ferenc-csaky: Got it. But I think doing this would complicate the code a lot, and we can simply apply an "add only if the user did not set it", which will only need to collect the keys the user explicitly set, and filter those out from whatever we set in `Sqrl2FlinkSQLTranslator`. WDYT?

mbroecheler: Yeah, you are right. The flow should probably be:
- get engine config
- overwrite mode
[Do all the parsing and planning]
- add any inferred config options when we generate the Physical plan at the very end and only set those if they aren't present already

And finally, specifically for the runner, we set options that we need only for the runner.

PR #1336, opened 2025-06-30
Feature: Argument Expressions and Custom SQL Queries

mbroecheler:
Adds support for two features:
- expressions in arguments
- sql query arguments

The former is not yet fully implemented and requires handling the parameter and instantiating the Flink function. The most work is going to be mapping from and to RowData. There is a test in DagPlannerTest called functionParameterExpressionTest.sqrl that is currently disabled because it doesn't work yet. Needs to be enabled.

The second is fully implemented but has poor test coverage at the moment.


ferenc-csaky:
Superseded by #1675

PR #1337, opened 2025-07-01
Add Postgres and Flink volumes, handle graceful Flink shutdown, add restore ability

ferenc-csaky:
## Main changes

> [!NOTE]
> I will clean-up the code a bit more, just wanted to get some feedback first.

- Added deafult Postgres volume as `/data/postgres` and made the necessary changes to init Postgres into that dir
- Added volumes to Flink checkpoints/savepoints
- Now `DatasqrlRun#run` can register a shutdown hook, that stops the Flink job with a savepoint that is saved under the configured savepoint dir
- Added a `-s, --savepoint` option to the `run` and `execute` commands, where when a savepoint (or checkpoint) path is specified, it restores the job from that.

## Arguable decision
Right now, the savepoint path can only come from the CLI and directly applied in `DatasqrlRun`, so anything that is specified for example in `package.json` will be overwritten. IMO that's not a problem, but maybe overlooked something.


sqrl-cli/Dockerfile.base
```diff
@@ -19,11 +21,8 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends redpanda \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
-# Setup postgres
-RUN service postgresql start \
-    && su - postgres -c "psql -U postgres -c \"ALTER USER postgres WITH PASSWORD 'postgres';\"" \
-    && su - postgres -c "psql -U postgres -c \"CREATE DATABASE datasqrl;\"" \
-    && su - postgres -c "psql -U postgres -c \"CREATE EXTENSION vector;\""
+COPY postgres.conf /etc/postgresql/${POSTGRES_VERSION}/main/conf.d/postgres.conf
+COPY redpanda.yaml /etc/redpanda/redpanda.yaml
```
velo: I would keep this file copy on the Dockerfile, cause these are more likely to change.

And COPY files is super fast too.

Goal when splitting the files was faster build, as buildx takes a long time to install dependencies.

sqrl-cli/src/main/java/com/datasqrl/cli/AbstractCompileCmd.java
```diff
@@ -92,7 +92,8 @@ protected void execute(ErrorCollector errors, Path snapshotPath, Optional<Path>
     var compilationProcess = injector.getInstance(CompilationProcess.class);
     testsPath.ifPresent(this::validateTestPath);
 
-    Pair<PhysicalPlan, ? extends TestPlan> plan = compilationProcess.executeCompilation(testsPath);
+    Pair<PhysicalPlan, ? extends TestPlan> plan =
```
velo: nit: use `var` when possible

sqrl-cli/src/main/java/com/datasqrl/cli/ExecuteCmd.java
```diff
@@ -29,10 +30,10 @@ public class ExecuteCmd extends AbstractCmd {
   protected void execute(ErrorCollector errors) throws Exception {
     var targetDir = getTargetDir();
     var planDir = targetDir.resolve(SqrlConstants.PLAN_DIR);
-    var sqrlConfig = ConfigLoaderUtils.loadPackageJson(targetDir);
+    var sqrlConfig = ConfigLoaderUtils.loadPackageJson(targetDir, ExecutionGoal.RUN);
```
mbroecheler: Here, we are loading the fully resolved and merged package config that comes out of compile. Hence, we should not be merging in anything anymore and only read the package json as is. As such, it should not be necessary to pass in the goal.

sqrl-cli/src/main/java/com/datasqrl/cli/RunCmd.java
```diff
@@ -39,10 +39,10 @@ public void execute(ErrorCollector errors) throws Exception {
     // Run
     var targetDir = getTargetDir();
     var planDir = targetDir.resolve(SqrlConstants.PLAN_DIR);
-    var sqrlConfig = ConfigLoaderUtils.loadPackageJson(targetDir);
+    var sqrlConfig = ConfigLoaderUtils.loadPackageJson(targetDir, getGoal());
```
mbroecheler: Same here

sqrl-planner/src/main/java/com/datasqrl/engine/stream/flink/FlinkStreamEngine.java
```diff
@@ -66,16 +65,15 @@ public ExecutionMode getExecutionMode() {
...
-  }
-
-  public Configuration getDirectoryConfig() {
+  public Configuration getBaseConfiguration() {
     var conf = new Configuration();
-    conf.set(CheckpointingOptions.CHECKPOINTS_DIRECTORY, "file:///data/flink/checkpoints");
-    conf.set(CheckpointingOptions.SAVEPOINT_DIRECTORY, "file:///data/flink/savepoints");
+    engineConfig.getConfig().forEach((key, value) -> conf.setString(key, String.valueOf(value)));
+
+    conf.set(
```
mbroecheler: I am wondering if we should get rid of our special "mode" configuration option and use Flink's instead. There is little value to have our own. I thought this was necessary since we need to set it in multiple places, but that was my misunderstanding.
This is not directly relevant to this particular PR but we can add it to the "config clean up" list.

ferenc-csaky: +1 to this, will include it to the change

ferenc-csaky:
> The direction looks good. It looks clean. Thank you. While we are at it, I am wondering if we can simplify all these "get config from package.json" methods we now have floating around. It is hard for me to discern them all ;-) Basically, we should have just a single static method in SqrlConfig for loading: loadConfig(ErrorCollector, Optional JsonSchema, JsonNode config) which validates the config if schema is present and initializes the SqrlConfig from it. The remaining logic for merging package files and resources and all that should move to ConfigLoaderUtils. There we should have two methods:
> 
>     1. one that takes a list of paths and a list of resources for defaults (for loading configs prior to compiling)
> 
>     2. one that takes a single path (for loading the resolved config from build dir)
>        The default values and lists and such should then live in the Packager which is responsible for the package.json management.
>        I think that would be a cleaner separation, hopefully clarify what goes where and eliminate redundancy. WDYT?

I like these suggestions and agree it would result in a much cleaner setup, will come up with another commit which incorporates all the comments/suggestions you made. Thanks for the thorough review and the extra thought!

mbroecheler (approved):
Thank you. I like how that refactoring landed.

PR #1340, opened 2025-07-02
Store snapshots as formatted files

velo:


mbroecheler (approved):
Nice way to make it backwards compatible.

PR #1345, opened 2025-07-04
Feature: Support batch mutations with transaction

mbroecheler:


sqrl-planner/src/main/java/com/datasqrl/planner/graphql/GraphqlSchemaValidator.java
```diff
@@ -304,13 +304,21 @@ private InputObjectTypeDefinition getValidMutationInputType(
...
 
+  private static Type unwrapNullAndList(Type type) {
+    for (; ; ) {
+      if (type instanceof NonNullType nullType) {
+        type = nullType.getType();
+      } else if (type instanceof ListType listType) {
+        type = listType.getType();
+      } else break;
+    }
+    return type;
```
velo: wondering if we could hit some bug that would make this hang....

maybe make it recursive, at least it would exit on stackoverflow

```
private static Type unwrapNullAndList(Type type) {
  if (type instanceof NonNullType nullType) {
    return unwrapNullAndList(nullType.getType());
  } else if (type instanceof ListType listType) {
    return unwrapNullAndList(listType.getType());
  } else {
    return type;
  }
}

```

sqrl-planner/src/main/java/com/datasqrl/util/EnumUtil.java
```diff
@@ -0,0 +1,29 @@
...
+package com.datasqrl.util;
+
+import java.util.Optional;
+
+public class EnumUtil {
+
+  public static <T extends Enum<T>> Optional<T> getByName(Class<T> enumClass, String name) {
+    try {
+      return Optional.of(Enum.valueOf(enumClass, name.trim().toUpperCase()));
+    } catch (IllegalArgumentException | NullPointerException e) {
```
velo: Better check for null instead of letting a NPE to be thrown

mbroecheler: I think Enum.valueOf() throws an exception. How would I get around that to check for null?

mbroecheler: Ah, I see. You already fixed it. Thank you

PR #1355, opened 2025-07-08
Fix container test exit code handling and log capture

velo:


sqrl-testing/sqrl-container-testing/pom.xml
```diff
@@ -81,6 +81,12 @@
       <artifactId>log4j-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <version>4.3.0</version>
```
ferenc-csaky: nit: let's add a version prop for this in the root POM

ferenc-csaky (approved):

PR #1356, opened 2025-07-09
Set random `PROPERTIES_GROUP_ID` if env variable has no value

ferenc-csaky:


ferenc-csaky:
> So, on the new design you don't need to change datasqrlTest anymore, right?

That's the idea, since inside `DatasqrlTest` also uses `DatasqrlRun`, so this should work, no matter how we end up in `DatasqrlRun`.

PR #1358, opened 2025-07-09
Bump Postgres to 17 and Flink to 1.19.3

ferenc-csaky:
Redpanda is installed from its stable repo, so currently it always installs the latest. `pgvector` is also installed via `postgresql-${POSTGRES_VERSION}-pgvector`, so `POSTGRES_VERSION` should cover that as well.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -124,9 +124,11 @@ All dependency versions should be centralized as properties in the root pom.xml 
 **Root POM Properties Location**: `/pom.xml` - All version properties are defined in the `<properties>` section.
 
 **Key Principles**:
-- NEVER use hardcoded versions in child module pom.xml files
-- ALWAYS add new version properties to the root pom.xml when introducing new dependencies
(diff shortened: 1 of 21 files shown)


sqrl-cli/Dockerfile.base
```diff
@@ -1,6 +1,6 @@
 FROM eclipse-temurin:17-jdk-noble
 
-ARG POSTGRES_VERSION=15
+ARG POSTGRES_VERSION=17
```
velo, line 3: 15 minutes to build the new base image... 15 minutes is crazy

ferenc-csaky: does the circleci build utilize docker cache? cause then it only will be this much once, until we change something, i guess

velo: well, kinda, using docker cache uses a lot of credits, but, the build hashes the Dockerfile.base, and check if there is an image that matches the hash.

If none is found it builds and pushes to docker hub

PR #1360, opened 2025-07-09
Configure container logging to write to /build/logs with console filtering

velo:
## Original Request

> can you change my application com.datasqrl.cli.DatasqrlCli to write all logs to `/build/logs` when I'm running it as a standalone container? I would like logs to be written for level INFO, I would like that each service I start on entrypoint to also write logs there too. And change the application to print logs only for WARN. You can keep com.datasqrl logs at INFO level.

## Summary

This PR implements comprehensive container logging configuration for the DataSQRL CLI application to centralize log output and optimize console verbosity.

### Changes Made

**1. Log4j2 Configuration (`sqrl-cli/src/main/resources/log4j2.properties`):**
- Added file appender to write all INFO+ level logs to `/build/logs/datasqrl-cli.log`
- Configured console logging to show only WARN+ level messages
- Created specific logger for `com.datasqrl` packages to maintain INFO level on console
- Added log rotation (daily + 10MB size limit, max 10 files)

**2. Container Entrypoint (`sqrl-cli/entrypoint.sh`):**
- Creates `/build/logs` directory at container startup
- Redirects Redpanda service logs to `/build/logs/redpanda.log`
- Redirects PostgreSQL service logs to `/build/logs/postgres.log`
- Updated ownership handling to fix permissions for entire `/build` directory

**3. Test Validation:**
- Enhanced `SqrlTestContainerIT` to validate log files are present and contain content
- Added `validateLogFiles` method to `SqrlContainerTestBase` for comprehensive log testing
- Validates CLI, Redpanda, and PostgreSQL log files exist with proper content

### Configuration Details

- **Console Output**: Only WARN+ level messages (except `com.datasqrl` packages show INFO)
- **File Logging**: All INFO+ level logs written to `/build/logs/`
- **Service Logs**: Redpanda and PostgreSQL services log to dedicated files
- **Log Rotation**: Files rotate daily or at 10MB, keeping max 10 files
- **Ownership**: Proper file ownership maintained for host system compatibility

### Test Results

✅ All tests pass - `SqrlTestContainerIT` successfully validates:
- Log directory creation
- Log file presence and content (CLI: 13KB, Redpanda: 331KB, PostgreSQL: 2.5KB)
- Console output filtering working correctly
- Service log capture functioning properly

The implementation ensures centralized logging while maintaining console readability and proper file ownership for container environments.

🤖 Generated with [Claude Code](https://claude.ai/code)


sqrl-cli/src/main/resources/log4j2.properties
```diff
@@ -15,16 +15,40 @@
...
+appender.file.fileName = /tmp/logs/datasqrl-cli.log
+appender.file.filePattern = /tmp/logs/datasqrl-cli-%d{yyyy-MM-dd}-%i.log
+appender.file.layout.type = PatternLayout
+appender.file.layout.pattern = %d{yyyy-MM-dd HH:mm:ss,SSS} [%t] %-5level %logger{36} - %msg%n
+appender.file.policies.type = Policies
+appender.file.policies.time.type = TimeBasedTriggeringPolicy
+appender.file.policies.size.type = SizeBasedTriggeringPolicy
+appender.file.policies.size.size=10MB
+appender.file.strategy.type = DefaultRolloverStrategy
+appender.file.strategy.max = 10
```
ferenc-csaky: 10MB seems a bit low, how about 50MB and keep 3 files?

sqrl-cli/entrypoint.sh
```diff
@@ -88,17 +100,22 @@ fi
 BUILD_UID=$(stat -c '%u' /build)
 BUILD_GID=$(stat -c '%g' /build)
 
-# Pre-create the build directory with proper ownership to avoid permission errors
+# Pre-create the build and logs directories with proper ownership to avoid permission errors
 mkdir -p /build/build 2>/dev/null || true
 chown -R "$BUILD_UID:$BUILD_GID" /build/ 2>/dev/null || true
 
+# Ensure logs directory has proper ownership
+chown -R "$BUILD_UID:$BUILD_GID" /tmp/logs 2>/dev/null || true
```
ferenc-csaky, line 108: I guess the reason why the logs go under `/tmp` cause it has a 777 permission set and anything can write into it, but that seems a hacky shortcut. :)

sqrl-cli/entrypoint.sh
```diff
@@ -16,8 +16,20 @@
 #
 
 set -e
+set -o pipefail
+
+# Enable debug mode if DEBUG environment variable is set
+if [ "$DEBUG" = "1" ]; then
```
ferenc-csaky: Single bracket conditions are an anti-pattern IMO, so we should avoid using it. 

Also, I think it does not matter what value `DEBUG` has, so if it is defined and not empty, we should `set -x`. For that, we could use a condition like this:
```bash
if [[ -n "${DEBUG+x}" && -n "$DEBUG" ]]; then
...
```

sqrl-cli/entrypoint.sh
velo: sqrl java code deletes /build/build when it starts.... so it delete the log folder after logs are create.... that is why I change to record at /tmp and move it later

ferenc-csaky: The part about not writing it under `/build` was clear, but logging under `/tmp` seems a bit odd to me, in most cases the target is under `/var/log` by default. For us that would mean `/var/log/sqrl/...`. But if that adds more complexity, `/tmp` is fine for now I guess, whatever happens, anything that is not mounted will be gone anyways.

sqrl-cli/src/main/resources/log4j2.properties
```diff
@@ -15,16 +15,40 @@
...
+appender.file.name = FileAppender
+appender.file.type = RollingFile
+appender.file.fileName = /tmp/logs/datasqrl-cli.log
+appender.file.filePattern = /tmp/logs/datasqrl-cli-%d{yyyy-MM-dd}-%i.log
+appender.file.layout.type = PatternLayout
+appender.file.layout.pattern = %d{yyyy-MM-dd HH:mm:ss,SSS} [%t] %-5level %logger{36} - %msg%n
+appender.file.policies.type = Policies
+appender.file.policies.time.type = TimeBasedTriggeringPolicy
+appender.file.policies.size.type = SizeBasedTriggeringPolicy
+appender.file.policies.size.size=50MB
```
ferenc-csaky: nit: missing whitespaces around `=`

sqrl-server/sqrl-server-vertx/entrypoint.sh
```diff
@@ -15,6 +15,11 @@
 # limitations under the License.
 #
 
+# Enable debug mode if DEBUG environment variable is set
+if [ "$DEBUG" = "1" ]; then
```
ferenc-csaky: Debug check can be updated here too

PR #1367, opened 2025-07-10
Add back AWS SDK to `sqrl-cli`

ferenc-csaky:
This was defined in the `sqrl-root` POM before, which i missed when i restructured the project.


velo (changes requested):
we have the same dependencies on integration tests modules, can we remove it from there?

velo (approved):
I will leave it as approved in case you need to move forward

PR #1376, opened 2025-07-14
Add a warning in compile time for the deprecated `name` key under `dependencies` in package config

ferenc-csaky:


mbroecheler (changes requested):
This change looks reasonable, but I find passing in the error collector into the configuration system a bit of an anti-pattern. SqrlConfig has an error collector already.
Hence, I would prefer if we add the ErrorCollector to DependencyImpl as a field and either get it from the SqrlConfig or pass it in explicitly (for test cases, using ErrorCollector.ROOT). Don't use null.

ferenc-csaky:
I ended up simplifying it even more, thanks for the comment, it made me think it through one more time. :)

mbroecheler (approved):

PR #1387, opened 2025-07-16
Rework dependent service start to be triggered only after compile

ferenc-csaky:
## Motivation
Starting the dependent services in `entrypoint.sh` meant that we has to start them before the `compile` phase, and in case of a `run` or `test`, if the compilation fails, we started them for nothing. Since now we only trigger 1 JVM process, that means to change this behavior, we need to start those from Java, during runtime.

## Main Changes
- Added `OsProcessManager` to start dependent services from the Java process, and manage directory ownership of container mounts.
- Added some extra logic to validate the proper startup of the dependent services
- Reworked env variable handling to use `System.getProperties` instead of `System.getenv`, cause properties can be modfied, and now `OsProcessManager` injects every env variable into the properties map

### DIFF
--- a/sqrl-cli/src/main/java/com/datasqrl/util/OsProcessManager.java
+++ b/sqrl-cli/src/main/java/com/datasqrl/util/OsProcessManager.java
@@ -75,7 +75,7 @@ public class OsProcessManager {
   public void teardown(Path buildDir) throws Exception {
     Path target = buildDir.resolve("logs");
     FileUtils.moveDirectory(Paths.get(LOGS_PATH).toFile(), target.toFile());
-    setOwnerForDir(buildDir);
+    setOwnerForDir(buildDir.getParent());
   }
 
   public void setOwnerForDir(Path dir) throws IOException, InterruptedException {
--- a/sqrl-cli/src/test/java/com/datasqrl/util/OsProcessManagerTest.java
+++ b/sqrl-cli/src/test/java/com/datasqrl/util/OsProcessManagerTest.java
@@ -442,6 +442,7 @@ class OsProcessManagerTest {
     when(mockBuildDir.resolve("logs")).thenReturn(mockTargetDir);
     when(mockBuildDir.toAbsolutePath()).thenReturn(mockBuildDir);
     when(mockBuildDir.toString()).thenReturn("/build/dir");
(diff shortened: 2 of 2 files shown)


sqrl-cli/entrypoint.sh
```diff
@@ -34,68 +34,6 @@ mkdir -p /tmp/logs
...
-
-    echo "All necessary services are up."
-
-    mkdir -p $FLINK_CP_DATA_PATH
-    mkdir -p $FLINK_SP_DATA_PATH
-fi
-
 # Get the UID/GID of the /build directory to ensure files are created with correct ownership
 BUILD_UID=$(stat -c '%u' /build)
 BUILD_GID=$(stat -c '%g' /build)
```
velo: we could also java the java app clearning up the `/build` ownership.... remove all from the script here and restore the `exec java ...`

sqrl-cli/src/main/java/com/datasqrl/cli/DependentServiceManager.java
```diff
@@ -0,0 +1,316 @@
...
+            "/etc/redpanda/redpanda.yaml",
+            "--smp",
+            "1",
+            "--memory",
+            "1G",
+            "--reserve-memory",
+            "0M",
+            "--node-id",
+            "0",
+            "--check=false");
```
velo: this would be easier to maintain if use a single string (probably a multiline `"""` and split on spaces.... it's same thing, but easier over time

sqrl-cli/entrypoint.sh
ferenc-csaky: To move that ownership mgmt to Java as well, I guess we would need to add a shutdown hook to set the ownership of the folder back

sqrl-cli/src/main/java/com/datasqrl/cli/DependentServiceManager.java
ferenc-csaky: I don't think that works, cause this is a vararg, and joining the whole thing messes up the `ProcessBuilder` AFAIK.

sqrl-cli/entrypoint.sh
velo: yes, or we can just run it before we trigger the system.exit, either is fine by me

PR #1392, opened 2025-07-17
Add `compiler/api` to the default configuration, standardize `enum` type fields in `packageSchema.json`

ferenc-csaky:


sqrl-planner/src/main/java/com/datasqrl/config/CompilerApiConfigImpl.java
```diff
@@ -16,58 +16,40 @@
...
 
-  /** How to generate the endpoints */
-  @Default Endpoints endpoints = Endpoints.FULL;
-
-  /** This suffix it appended to all table names (before the uid) */
-  @Default boolean addOperationsPrefix = true;
+  @Override
+  public EnumSet<ApiProtocol> getProtocols() {
+    return EnumSet.copyOf(sqrlConfig.asList("protocols", ApiProtocol.class).get());
+  }
```
velo, line 31: make sense, move all the logic from here to default-package.json

sqrl-planner/src/test/java/com/datasqrl/config/CompilerConfigImplTest.java
```diff
@@ -50,7 +50,19 @@ void givenCostModel_whenInvalid_thenThrowsError() {
...
         .isInstanceOf(CollectedException.class)
-        .hasMessage(
-            "Value [asd] for key [cost-model] is not valid. Must be one of [DEFAULT, READ, WRITE]");
+        .hasMessageContaining(
+            "String \"asd\": not one of the values accepted for Enum class: [DEFAULT, READ, WRITE]");
+  }
+
+  @Test
+  void givenCostModel_whenLowercase_thenThrowsError() {
+    config.setProperty("cost-model", "read");
```
velo, line 59: sqrl is extremely case insensitive in most cases.

We even go out of our way to locate files ignoring case on case sensitive file systems

sqrl-planner/src/test/java/com/datasqrl/config/PackageJsonSchemaTest.java
```diff
@@ -55,19 +55,16 @@ void validConfigFile(String configFileName) {
       strings = {
         "emptyEnabledEngines.json",
         "invalidVersionFormat.json",
-        "additionalPropertyInIceberg.json",
-        "invalidConnectorInIceberg.json",
```
velo, line 59: curious to know why these got removed

velo (approved):

sqrl-planner/src/test/java/com/datasqrl/config/PackageJsonSchemaTest.java
ferenc-csaky: Cause we do not validate these in `packageSchema.json`. It just did not show up, cause the test case that expected to throw an error failed to find the JSON files, cause the path was configured wrong, hence it passed for all of the until now.

PR #1397, opened 2025-07-18
Add thread-safe `GlobalEnvironmentStore` to handle and access env variables

ferenc-csaky:


sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/env/GlobalEnvironmentStore.java
```diff
@@ -0,0 +1,136 @@
...
+ * environment variables throughout the JVM process.
+ *
+ * <p>This class uses Guava's Cache for optimal thread-safe read performance and is designed to be
+ * accessed from any module in the DataSQRL project.
+ */
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@Slf4j
+public final class GlobalEnvironmentStore {
+
+  private static final Cache<String, String> cache = CacheBuilder.newBuilder().build();
```
velo, line 36: Shouldn't this be a loading cache that gets initialized with System.getenv ?

sqrl-cli/src/main/java/com/datasqrl/util/OsProcessManager.java
```diff
@@ -157,9 +158,9 @@ private void startRedpanda() throws IOException, InterruptedException {
     waitForRedpanda("localhost", 9092);
 
     // Set environment variables
-    setSystemProperty("KAFKA_HOST", "localhost");
-    setSystemProperty("KAFKA_PORT", "9092");
-    setSystemProperty("PROPERTIES_BOOTSTRAP_SERVERS", "localhost:9092");
+    setEnvironmentVariable("KAFKA_HOST", "localhost");
+    setEnvironmentVariable("KAFKA_PORT", "9092");
+    setEnvironmentVariable("PROPERTIES_BOOTSTRAP_SERVERS", "localhost:9092");
```
velo, line 163: The problem I see here, is that I'm no longer able to override `PROPERTIES_BOOTSTRAP_SERVERS`

It was possible in previous versions of sqrl and we need it again

ferenc-csaky: Override how? Anything that is passed at container start via `-e` will be added to the `GlobalEnvironmentStore`. Anything we wanna change during runtime, we can change it via `GlobalEnvironmentStore.put(key, value)`. I may miss something, but I am not aware where did we override env variables and how precisely.

velo: So, I start a container and I set `-e PROPERTIES_BOOTSTRAP_SERVERS=remoteserver:1234`

setEnvironmentVariable will delete ignore my custom server

ferenc-csaky: If you wanna bring Kafka from another container `KAFKA_HOST` should be provided, cause that is the env.var that is checked to start an internal redpanda or not. We can change that to `PROPERTIES_BOOTSTRAP_SERVERS`, I just kept using the one we used previously in `entrypoint.sh`.

But even if an internal redpanda is started, if an environment variable was already present before the Java app start, it won't be overridden, cause i add `System.getenv()` to the `GlobalEnvironmentStore` AFTER all the dependent services started, asd the last step [here](https://github.com/DataSQRL/sqrl/blob/5b6df7f62b87a95b1ad56b71a6754c0dc371fcb4/sqrl-cli/src/main/java/com/datasqrl/util/OsProcessManager.java#L64).

velo: ok, lemme try that then

velo (approved):

PR #1402, opened 2025-07-21
Add ability to run outside projects via `FullUseCasesIT`

ferenc-csaky:


sqrl-testing/sqrl-integration-tests/src/test/java/com/datasqrl/FullUseCasesIT.java
```diff
@@ -334,64 +336,70 @@ void printUseCaseNumbers(UseCaseTestParameter param) {
   @Test
   @Disabled
   public void runTestCaseByName() {
```
velo, line 339: I would prefer having a new test just for external projects....

velo (approved):

sqrl-testing/sqrl-integration-tests/src/test/java/com/datasqrl/FullUseCasesIT.java
ferenc-csaky: That's where i started as well, and ended up with some duplicates so merged them, but lets see, i'll give it another spin and abstract the common parts

PR #1407, opened 2025-07-22
fix(vertx): Resolve uncompleted promise in async operations

velo:


sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/GraphQLServerVerticle.java
```diff
@@ -104,7 +104,9 @@ protected void setupGraphQLRoutes(Promise<Void> startPromise) {
         .handler(GraphQLWSHandler.create(this.graphQLEngine))
         .handler(GraphQLHandler.create(this.graphQLEngine, this.config.getGraphQLHandlerOptions()));
 
-    startPromise.complete();
+    if (model.getSubscriptions().isEmpty()) {
+      startPromise.complete();
+    }
```
velo, line 109: When we have subscriptions, we complete the startPromise here
https://github.com/DataSQRL/sqrl/pull/1407/files#diff-d8eec2d1c07bf603e1cd091a58acda905a7e7c8df72295153d56c04d3234500bR77

PR #1412, opened 2025-07-22
Add container tests for HTTP request tracing functionality

velo:
Add test to verify `DATASQRL_TRACE_REQUESTS=true` logs GraphQL request bodies

### DIFF
--- a/sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/HttpServerVerticle.java
+++ b/sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/HttpServerVerticle.java
@@ -108,13 +108,6 @@ public class HttpServerVerticle extends AbstractVerticle {
   private void bootstrap(Promise<Void> startPromise) {
     Router root = Router.router(vertx);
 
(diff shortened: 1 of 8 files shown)


sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/HttpServerVerticle.java
```diff
@@ -129,6 +122,13 @@ private void bootstrap(Promise<Void> startPromise) {
     root.route().handler(toCorsHandler(config.getCorsHandlerOptions()));
     root.route().handler(BodyHandler.create());
 
+    // Use detailed tracing if enabled, otherwise use standard logging (must be after BodyHandler)
+    if (Boolean.parseBoolean(System.getenv().getOrDefault("DATASQRL_TRACE_REQUESTS", "false"))) {
```
ferenc-csaky: nit: I know it was already like this but how about `SQRL_TRACE_REQUESTS`? It's shorter and i also introduce a bunch of new internal env variables with that prefix, so we could use a consistent prefix.

PR #1418, opened 2025-07-23
Standardize env var naming, start dependent services dynamically, based on plan

ferenc-csaky:
## Main changes
- Standardized env variable naming: added dependent service prefix (`KAFKA_`, `POSTGRES_`), and applied some rewording.
- Detect dynamically what service should we start based on the appropriate plan files.
- The `kafka.json` and `postgres.json` content loading were already had an implementation in `DatasqrlRun`, and now that we use that in multiple places, I extracted those to `ConfigLoaderUtils`.
- Removed every test-specific logic from `DatasqrlRun` and now whatever we need for test running it is set up by `TestExecutionEnv` (added `loadInternalTestFlinkConfig` there).

## ~TODO~
- [x] Document what env vars we use internally.


sqrl-cli/src/main/java/com/datasqrl/cli/DatasqrlRun.java
```diff
@@ -188,68 +175,52 @@ private TableResult runFlinkJob() {
 
   @SneakyThrows
   private void initKafka() {
-    if (!planDir.resolve("kafka.json").toFile().exists()) {
+    var topics = ConfigLoaderUtils.loadKafkaTopics(planDir);
+    if (topics.isEmpty()) {
+      log.warn(
```
mbroecheler: we should still create the topics if only create-topics are specified. This is needed when using run command to create customer specific test cases, for example (see slack comment)

sqrl-cli/src/main/java/com/datasqrl/cli/DatasqrlRun.java
```diff
@@ -188,68 +175,52 @@ private TableResult runFlinkJob() {
...
-    }
+    var mutableTopics = new ArrayList<>(topics);
+    mutableTopics.addAll(sqrlConfig.getTestConfig().getCreateTopics());
 
-    List<Map<String, Object>> mutableTopics = new ArrayList<>(topics);
-    sqrlConfig
-        .getTestConfig()
-        .getCreateTopics()
-        .forEach(topic -> mutableTopics.add(Map.of("topicName", topic)));
+    var bootstrapServers = getenv("KAFKA_BOOTSTRAP_SERVERS");
```
mbroecheler: Can we make these env vars static strings is some class so they are all in one place. Make management and maintenance easier (if we need to rename them again).

sqrl-cli/src/main/java/com/datasqrl/util/OsProcessManager.java
```diff
@@ -105,67 +139,82 @@ private void createDirectoryWithPermission(Path dir) throws IOException, Interru
...
 
-    waitForRedpanda("localhost", 9092);
-
     // Set environment variables
-    setEnvironmentVariable("KAFKA_HOST", "localhost");
-    setEnvironmentVariable("KAFKA_PORT", "9092");
-    setEnvironmentVariable("PROPERTIES_BOOTSTRAP_SERVERS", "localhost:9092");
-
-    log.info("Redpanda started successfully");
+    setEnvironmentVariable("KAFKA_BOOTSTRAP_SERVERS", bootstrapServers);
```
mbroecheler: And then we can reference the same static variable here.

sqrl-cli/src/main/java/com/datasqrl/util/OsProcessManager.java
```diff
@@ -203,33 +252,34 @@ private void startPostgres() throws IOException, InterruptedException {
...
-      setEnvironmentVariable("PGPORT", "5432");
-      setEnvironmentVariable("PGPASSWORD", "postgres");
-      setEnvironmentVariable("PGDATABASE", "datasqrl");
+    if (!started) {
+      log.info("Starting Postgres service ...");
+      startPostgresService();
     }
 
+    // Set environment variables
+    setEnvironmentVariable("POSTGRES_HOST", LOCALHOST);
```
mbroecheler: and here

sqrl-cli/src/main/java/com/datasqrl/util/OsProcessManager.java
```diff
@@ -203,33 +252,34 @@ private void startPostgres() throws IOException, InterruptedException {
...
+    setEnvironmentVariable("POSTGRES_AUTHORITY", PG_AUTHORITY);
+    setEnvironmentVariable("POSTGRES_JDBC_URL", "jdbc:postgresql://" + PG_AUTHORITY);
+    setEnvironmentVariable("POSTGRES_USERNAME", "postgres");
+    setEnvironmentVariable("POSTGRES_PASSWORD", "postgres");
+
     log.info("Postgres started successfully");
   }
 
+  private boolean isKafkaPlanned(Path planDir) {
+    return !ConfigLoaderUtils.loadKafkaTopics(planDir).isEmpty();
```
mbroecheler: this needs to account for create-topics in package.json

sqrl-cli/src/main/java/com/datasqrl/cli/DatasqrlRun.java
mbroecheler: Actually, a different way to go about this that may be less "magical": We add the create-topics to the KafkaPhysicalPlan so they are part of the kafka.json. Then it's more obvious to the user.
In KafkaLogEngine, at the very end of the class we create a list of all NewTopic. We are already injecting a PackageJson into the class anyways. Should be a pretty simple addition and you can check if a topic name is duplicated and raise an error.

sqrl-planner/src/main/java/com/datasqrl/util/ConfigLoaderUtils.java
```diff
@@ -198,6 +202,68 @@ public static Configuration loadFlinkConfig(Path planDir) throws IOException {
...
+  /**
+   * Loads Kafka topic names from the plan directory by parsing the {@code kafka.json} configuration
+   * file.
+   *
+   * @param planDir the plan directory containing the {@code kafka.json} file
+   * @return an unmodifiable list of topic names found in the configuration, or an empty list if no
+   *     {@code kafka.json} file exists or if parsing fails
+   * @throws IllegalArgumentException if the plan directory is null or does not exist
+   */
+  public static List<String> loadKafkaTopics(Path planDir) {
```
mbroecheler: Let's parse this back into NewTopic. Currently, we only need the name but in the future we want to extend this to include other configuration parameters like compaction or TTL.

mbroecheler (changes requested):
Overall this looks great. Some minor suggestions for topic handling.

sqrl-planner/src/main/java/com/datasqrl/util/ConfigLoaderUtils.java
```diff
@@ -198,6 +204,81 @@ public static Configuration loadFlinkConfig(Path planDir) throws IOException {
...
+   * @throws IllegalArgumentException if the plan directory is null or does not exist
+   */
+  public static KafkaPhysicalPlan loadKafkaPhysicalPlan(Path planDir) {
+    validatePlanDir(planDir);
+
+    List<NewTopic> topics = List.of();
+    Set<String> testRunnerTopics = Set.of();
+
+    var kafkaFile = planDir.resolve("kafka.json").toFile();
+    if (kafkaFile.exists()) {
```
mbroecheler, line 220: Why is all this code necessary? Can you just deserialize the KafkaPhysicalPlan directly with the object mapper? You might have to set a no-arg constructor or such on the plan. But it should be simple to de/serialize.

sqrl-planner/src/main/java/com/datasqrl/engine/log/kafka/KafkaPhysicalPlan.java
```diff
@@ -16,11 +16,19 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import java.util.List;
+import java.util.Set;
 import lombok.Value;
 
 @Value
 public class KafkaPhysicalPlan implements EnginePhysicalPlan {
 
   List<NewTopic> topics;
+  Set<String> testRunnerTopics;
```
mbroecheler: imho this should be another List<NewTopic>. It makes sense to break out the test runner topics but I think it will be more future proof to use the same class. Then you can add a helper method to return all topics for the test runner.

PR #1423, opened 2025-07-24
Fix Swagger REST endpoint paths and add operation descriptions

velo:
## Summary
- Fix Swagger generated paths to include `/rest` prefix matching actual server endpoints  
- Use function descriptions from DataSQRL operations instead of generic auto-generated text

## Changes
- Updated `SwaggerService` constructor to accept REST endpoint configuration
- Modified path generation to prepend REST endpoint prefix  
- Extract operation descriptions from `FunctionDefinition` objects
- Fallback to generic description when no description available

Addresses PR feedback comments about incorrect base URLs and missing operation descriptions.


ferenc-csaky (approved):
LGTM

mbroecheler (approved):

PR #1427, opened 2025-07-25
Add blocking after Flink job `FINISHED` in case of `run` or `execute`

### LINKED ISSUE #1426: Keep the Docker container running for `run` and `execute` eve if the Flink job gets FINISHED
Since we also can have Postgres and Vert.x that should be able to serve data even in case a bounded Flink job terminated gracefully.

ferenc-csaky:
Resolves #1426


mbroecheler (approved):
thank you ;-)

PR #1437, opened 2025-07-29
Introduce `AbstractFullUseCaseTest` to bring `FullUseCaseIT` and `ExternalUseCaseIT` to a common parent layer

ferenc-csaky:
## Main changes
- Move the common test logic from `FullUseCasesIT` to `AbstractFullUseCaseTest`
- Move `TestScriptCompileExtension` logic to `AbstractFullUseCaseTest` and delete extension
- Invoke `DatasqrlTest` directly in `AbstractFullUseCaseTest`
- Delete `TestExecutionEnv` (invoked `DatasqrlTest` previously, now it is unused)
- Use the new abstract class in `FullUseCaseIT` (renamed)
- Use the new abstract class in `ExternalUseCaseIT` (renamed)


sqrl-testing/sqrl-integration-tests/src/test/java/com/datasqrl/ExternalUseCaseIT.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.stream.Stream;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+/** Tests external use cases manually, provided via {@code externalUseCaseProvider}. */
+@Slf4j
+@Disabled
+public class ExternalUseCaseIT extends AbstractFullUseCaseTest {
```
velo, line 30: Wouldn't be better to make this a `public static void main`? 

Then we don't have this enable disable test issue

sqrl-testing/sqrl-integration-tests/src/test/java/com/datasqrl/ExternalUseCaseIT.java
```diff
@@ -0,0 +1,54 @@
...
+  @MethodSource("externalUseCaseProvider")
+  void testCase(UseCaseParam param) {
+    fullUseCaseTest(param);
+  }
+
+  static Stream<UseCaseParam> externalUseCaseProvider() {
+    // Provide absolute paths for the package.json file of the external project(s) to test
+    Stream<Path> pathStream =
+        Stream.of(
+            // Path.of("<absolute-path-of-external-package-json>")
```
velo, line 42: so, you saying I can just point to my local copy of https://github.com/DataSQRL/cloud-backend/blob/main/core/package-test-static.json and it would work

THIS IS AWESOME

velo (approved):

sqrl-testing/sqrl-integration-tests/src/test/java/com/datasqrl/ExternalUseCaseIT.java
ferenc-csaky: correctm, i tested it with `datasqrl-examples` projects. it's not going to every length though, cause if the some UDF classes are loaded, there is no way to debug that code ATM, but that's pretty advanced stuff on its own :D

sqrl-testing/sqrl-integration-tests/src/test/java/com/datasqrl/ExternalUseCaseIT.java
ferenc-csaky: it's mandatory to get this spinned up via JUnit, cause `AbstractFullUseCaseTest` has JUnit-specific setup/teardown logic, which sets up the test environment

PR #1440, opened 2025-07-30
Relocate `flinkrunner` classes internally

ferenc-csaky:
This is crucial to make sure we do not end up with classloader clashes when users may shade `flinkrunner` classes into their UDF JARs.


velo (approved):
For now is fine, but we should probably invest in better classloading that isolate classes and stop collisions
https://github.com/DataSQRL/cloud-compilation/blob/main/cloud-compilation/src/main/java/com/datasqrl/kubedeploy/steps/compilation/SqrlBridge.java#L39

Cause udfs can include a number of dependencies that would conflict with sqrl dependencies.

Honestly, the better fix for this would be us publishing a bom artifact that make sure all sqrl, flink-sql-runner and it's dependencies as provided dependencies

PR #1447, opened 2025-08-01
Refactor schema import, script import, and packager

mbroecheler:
This PR contains breaking changes.

Motivation:
One thing that is interesting about working with AI/LLMs to generate SQRL files/projects is seeing where it trips up. Two things it has a lot of trouble with:
1. Using schema files as the basis for CREATE TABLE statements using the file convention that we have in SQRL (e.g. table-name.avsc and table-name.table.json)
2. Using dependencies to manage multiple variants of table sources

What’s interesting about that is those are 2 things that SQRL does slightly differently from other programming languages and hence is “not intuitive”. In fact, I ran this by our customers at Medidata and they had trouble understanding this as well.
So, I’m working on an update for 0.8 to make this more intuitive:

1. To pull in an external schema into a table definition, you have to explicitly reference the file in the LIKE statement of the table. We “extend” the LIKE functionality that FlinkSQL provides so that if the string in the LIKE clause references an external file and it is a recognized schema file, we use that to create the base table.
2. I suggest we move to mustache.java (or any other templating engine for that matter) to do variable substitution on SQRL files for handling the dynamic elements of SQRL instead of using our dependency management. It’s more flexible and more obvious. So, instead of having an IMPORT my-datasource.my-table;  statement where my-datasource isn’t a folder that actually exists but is mapped in the dependencies to the folder my-datasource-kafka and the user has to know this and look at the package json to figure this out, we will have IMPORT my-datasource.my-table-{{variant}};where variant is a variable that is defined in the package.json to identify which table source i want to import. This makes it more obvious since I can look in my local filesystem and see my-table-kafka.table.json and my-table-test.table.json and realize that those are the variants that I can reference. Less guessing.
3. While I’m at it, I’m also extending our import mechanism to import scripts into separate databases so we don’t have to hide tables so much. It will look prettier and provide better isolation.

TODO:
- [x] initial refactoring
- [x] updating tests
- [x] adding test coverage
- [x] updating documentation
- [x] updating datasqrl-examples (create separate ticket)

This requires a major release due to breaking changes.


ferenc-csaky:
@mbroecheler I already like the added/deleted line ratio :)

ferenc-csaky (approved):

PR #1452, opened 2025-08-05
Make DuckDB and Snowflake query clients part of `server-config.json`

ferenc-csaky:
## Main changes
- Add DuckDB and Snowflake JDBC related options we currently set to `server-config.json`
- Add `JdbcConfig` to handle DuckDB and Snowflake config options and do the relevant mappings in the `ServerConfig` POJO construction
- Refactor `JdbcClientsConfig` to use `ServerConfig` for every client initialization
- Detect enabled query engines in `GenericJavaServerEngine` and disable query engines according to `package.json`
- Some trivial `class` -> `record` refactor
- Enable `duckdb` usecase in `FullUseCaseIT`, after adding the proper `--add-opens` to the `maven-failsafe-plugin` it runs properly
- Change `url` setup in Snowflake engine to not hardcode anything, but use 1 simple env var `SNOWFLAKE_JDBC_URL`
- Add Snowflake env vars to `configuration.md`


pom.xml
```diff
@@ -955,6 +955,13 @@
...
             <flinkrunner.version>${flinkrunner.version}</flinkrunner.version>
           </systemPropertyVariables>
+          <argLine>${argLine}
+            --add-opens=java.base/java.util=ALL-UNNAMED
+            --add-opens=java.base/java.lang=ALL-UNNAMED
+            --add-opens=java.base/java.io=ALL-UNNAMED
+            --add-opens=java.base/java.nio=ALL-UNNAMED
+            --add-opens=java.base/java.net=ALL-UNNAMED
+          </argLine>
         </configuration>
```
velo, line 965: I would prefer if you moved this settings to
https://github.com/DataSQRL/sqrl/blob/main/.mvn/jvm.config

sqrl-cli/src/main/resources/templates/server-config.json
```diff
@@ -195,5 +195,13 @@
     "value.deserializer" : "com.datasqrl.graphql.kafka.JsonDeserializer",
     "bootstrap.servers" : "${KAFKA_BOOTSTRAP_SERVERS}",
     "auto.offset.reset" : "latest"
+  },
+  "duckDbConfig": {
+    "driver": "org.duckdb.DuckDBDriver",
+    "url": "jdbc:duckdb:"
```
velo: Feels like you are missing an env var here?!

sqrl-planner/src/main/java/com/datasqrl/engine/server/VertxEngineFactory.java
```diff
@@ -40,7 +40,7 @@ public static class VertxEngine extends GenericJavaServerEngine {
 
     @Inject
     public VertxEngine(PackageJson packageJson) {
-      super(ENGINE_NAME, packageJson.getEngines().getEngineConfigOrEmpty(ENGINE_NAME));
+      super(ENGINE_NAME, packageJson.getEngines().getEngineConfigOrEmpty(ENGINE_NAME), packageJson);
```
velo: Wondering if we should still be performing the `packageJson.getEngines().getEngineConfigOrEmpty` here if we are going to pass the whole `packageJson`


I would say to either just inject `packageJson` OR replace `packageJson` with `packageJson.getEnabledEngines()`

sqrl-planner/src/main/resources/default-package.json
```diff
@@ -38,7 +38,7 @@
       "schema-type": "aws-glue",
       "catalog-name": "${SNOWFLAKE_CATALOG_NAME}",
       "external-volume": "${SNOWFLAKE_EXTERNAL_VOLUME}",
-      "url": "jdbc:snowflake://${SNOWFLAKE_ID}.snowflakecomputing.com/?user=${SNOWFLAKE_USER}&password=${SNOWFLAKE_PASSWORD}&warehouse=COMPUTE_WH&db=MYSNOWFLAKEDB&schema=public&disableSslHostnameVerification=true"
+      "url": "${SNOWFLAKE_JDBC_URL}"
```
velo, line 41: much better :+1:

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/jdbc/PreparedSqrlQueryImpl.java
```diff
@@ -17,11 +17,6 @@
...
-import lombok.Value;
 
-@Value
-public class PreparedSqrlQueryImpl implements PreparedSqrlQuery<String> {
-
-  Connection connection;
-  String preparedQuery;
-}
+public record PreparedSqrlQueryImpl(Connection connection, String preparedQuery)
+    implements PreparedSqrlQuery<String> {}
```
velo, line 22: :heart: 

I hate `@Value`, glad you are not a big fan of it either =)

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/jdbc/JdbcExecutionContext.java
```diff
@@ -79,14 +79,13 @@ public CompletableFuture runQuery(ResolvedSqlQuery resolvedQuery, boolean isList
...
-                    .prepareStatement(preparedQueryContainer.getPreparedQuery())) {
+            try (final Connection conn = preparedQueryContainer.connection();
+                final PreparedStatement stmt =
+                    conn.prepareStatement(preparedQueryContainer.preparedQuery())) {
               for (int i = 0; i < paramObj.size(); i++) {
-                statement.setObject(i + 1, paramObj.get(i));
+                stmt.setObject(i + 1, paramObj.get(i));
               }
-              ResultSet resultSet = statement.executeQuery();
+              ResultSet resultSet = stmt.executeQuery();
```
velo: no need to change it now, but, I would prefer to see more `var`s as we refactor code

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/JdbcClientsConfig.java
```diff
@@ -17,94 +17,80 @@
...
-  private SqlClient createPostgresSqlClient() {
+  private SqlClient initPostgresSqlClient() {
     var poolOptions = new PoolOptions(this.config.getPoolOptions());
     // Note: setPipelined() method was removed in Vert.x 5, pipelining is now always enabled
     return Pool.pool(vertx, this.config.getPgConnectOptions(), poolOptions);
   }
+
+  @SneakyThrows
+  private Pool initJdbcPool(String driver, String url) {
+    Class.forName(driver);
```
velo: I wonder if we could turn this into a compile time check.

Instead of `initJdbcPool(duckDbConf.getDriver(), url);` have `initJdbcPool(org.duckdb.DuckDBDriver.class, url);`

Does it make sense to make jdbc driver configurable?!

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/VertxJdbcClient.java
```diff
@@ -29,16 +29,12 @@
 import io.vertx.sqlclient.SqlClient;
 import io.vertx.sqlclient.Tuple;
 import java.util.Map;
-import lombok.Value;
 
 /**
  * Purpose: Manages SQL clients and executes queries. Collaboration: Used by {@link VertxContext} to
  * prepare and execute SQL queries.
  */
-@Value
```
velo, line 38: Probably for 0.8 we should tell claude.

```
Hi, please review all `@Value` usage.

For any classes that contain less then 3 methods convert to `java record`
```

And let it run wild

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/config/ServerConfig.java
```diff
@@ -47,29 +47,32 @@ public class ServerConfig {
   SwaggerConfig swaggerConfig;
   KafkaConfig.KafkaMutationConfig kafkaMutationConfig;
   KafkaConfig.KafkaSubscriptionConfig kafkaSubscriptionConfig;
+  JdbcConfig duckDbConfig;
+  JdbcConfig snowflakeConfig;
 
   public ServerConfig(JsonObject json) {
```
velo, line 53: This parser is so complicated, so complicated that I wonder if we should dump it write generic java records and use mapstrut to translate our objects into vertx objects.

This parser feels like 2004 hibernate

velo (approved):

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/config/ServerConfig.java
ferenc-csaky: Couldn't agree more, lets get that done in 0.8

sqrl-planner/src/main/java/com/datasqrl/engine/server/VertxEngineFactory.java
ferenc-csaky: My general idea was to pass thorugh `duckdb` and `snowflake` configs from `package.json`, cause that would be the best UX for customers, but for this specific change that seemed too much change, so i'd rather do something like that in the next minor release.

So i added the whole `packageJson` here for future usability, but for now, u're right that like this it's not the best.

pom.xml
ferenc-csaky: `.mvn/jvm.config` won't picked up in the CI run for the failsafe plugin, that's why i added here

sqrl-cli/src/main/resources/templates/server-config.json
ferenc-csaky: nope, `jdbc:duckdb:` is the valid in-memory "url" for duckdb, it was the same before. did not add user configurability at this point

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/JdbcClientsConfig.java
ferenc-csaky: actually those are not needed at all, cause both jdbc driver are available compile-time and gets auto-registered

PR #1457, opened 2025-08-06
Adding cache and ttl hints

mbroecheler:
This PR adds support for time-to-live and cache duration hints in SQRL.
The ttl applies to the tables and topics created.
The cache duration applies to queries executed against the database.

This PR is not yet complete. See AbstractJdbcStatementFactory#createTable for the remaining implementation TODOs.
In particular, we need to figure out how to refactor this class to make the CREATE TABLE ddl adjust to each dialect.

Also, snapshots need to be updated.

This PR does not implement the caching on the vertx server. That requires follow-on work (create ticket once this is complete).


mbroecheler:
Refactored the table handling and generalized our partitioning model according to:
https://docs.google.com/document/d/127xDC9S3wwizvweM8OU94zu8vXnTxukvt_PfQpG0Hvw/edit?tab=t.0

This fixes the remaining issues and puts us on a solid footing regarding partitioning in the various engines we support (and want to support) with support for optimization down the road.

What is missing:
- a use case test that tests postgres range and hash partitioning (see DagPlannerTest#partitionTest for what this should look like)
- fixing snapshots

ferenc-csaky (approved):

PR #1462, opened 2025-08-06
Resolve a couple TODOs

ferenc-csaky:


sqrl-cli/src/main/java/com/datasqrl/cli/DatasqrlTest.java
```diff
@@ -60,14 +51,14 @@
  */
 public class DatasqrlTest {
 
+  private static final String GRAPHQL_ENDPOINT = "http://localhost:8888/graphql";
+  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
```
velo: there is a DatasqrlObjectMapper, use that cause we need to register some module for this to work well.

sqrl-cli/src/main/java/com/datasqrl/compile/GqlGenerator.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import lombok.RequiredArgsConstructor;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+@RequiredArgsConstructor
+class GqlGenerator extends GraphqlSchemaVisitor {
```
velo, line 51: What Gql means?

ferenc-csaky: shorthand for grapghql i guess, i not created this, just moved it.

PR #1463, opened 2025-08-06
Separate Vert.x debug log config and use it only in DEBUG mode

ferenc-csaky:


sqrl-server/sqrl-server-vertx/entrypoint.sh
```diff
@@ -15,13 +15,16 @@
 # limitations under the License.
 #
 
+SQRL_JVM_ARGS=
+
 # Enable debug mode if DEBUG environment variable is set
 if [[ -n "${DEBUG+x}" && -n "$DEBUG" ]]; then
+    SQRL_JVM_ARGS="-Dlog4j2.configurationFile=/opt/sqrl/app/log4j2-debug.properties"
```
velo, line 22: in that case, can we remove `SQRL_TRACE_REQUESTS` , otherwise we need to enable 2 flags

velo (approved):

PR #1467, opened 2025-08-07
Validate and snapshot any non GraphQL query JDBC views in `DatasqrlTest`

ferenc-csaky:


mbroecheler (changes requested):
This looks great. My thinking is that we should NOT snapshot views by default but only run the query and make sure there are no exceptions.
It can be confusing to users to have a bunch of snapshots without the ability to control what gets snapshot. If they want to snapshot a view, they can easily add a simple `SELECT * FROM MyView;` test for it.
My recommendation is to only execute the query without snapshotting and then you don't have to filter out certain views that are covered by tests. Makes things simpler and more predictable.

mbroecheler (approved):
minor nit: queryAndSnapshotJdbcViews -> validateJdbcViews

PR #1478, opened 2025-08-12
Map query engine JDBC urls from the package config

ferenc-csaky:


sqrl-planner/src/main/java/com/datasqrl/engine/server/VertxEngineFactory.java
```diff
@@ -40,7 +46,45 @@ public static class VertxEngine extends GenericJavaServerEngine {
 
     @Inject
     public VertxEngine(PackageJson packageJson) {
-      super(ENGINE_NAME, packageJson);
+      super(ENGINE_NAME, packageJson, getQueryEngineConfigs(packageJson));
```
mbroecheler, line 49: It looks like we can make this code DRYer

mbroecheler: In fact, I wonder if we can generalize this: if all we need on the server side for query engines in the URL, I wonder if we just extract all the query engines that are defined in the package.json and move their URL over. Then this will generalize for additional query engines like Athena, Trinio, etc

ferenc-csaky: Completely agree. But for 0.8 I really want to refactor how the JSON handling works on the server side and throw away that vert.x `JsonObject` as far as I can, so I did not want to go with this too. Until then I think referencing snowflake and duckdb explicitly here is tolerable. I'll add a TODO to the code and an issue for 0.8.

mbroecheler: Cool, if this is just a stopgap until we can refactor this I'm good.

PR #1487, opened 2025-08-13
Fix Flink FS integration, add S3 sink container test

ferenc-csaky:
## Main changes
- Changed `hadoop.version` to the exact same version Flink 1.19 defines for the `fs` modules, otherwise if we run into a Hadoop FS class implementation for some reason, reading the default config fails.
- Added `org.apache.flink.core.fs.FileSystem.initialize(...)` call in `DatasqrlRun` to properly initialize the Flink FS layer.
- Created a container test that writes a JSON file into a minio S3 bucket so if this breaks at some point again we will know.

## TODO
- [x] Validate if `FileSystem.initialize` also required on the `flink-sql-runner` side, cause if yes, I move it there and then it can be omitted in this repo.


ferenc-csaky:
I think the `FileSystem#initialize` call should be moved to `flink-sql-runner`, opened a PR: https://github.com/DataSQRL/flink-sql-runner/pull/185

But that will require a new runner release, so in the meantime, we can merge this in, and then the init can be removed when we bump the runner version.

PR #1510, opened 2025-08-25
Support GraphQL API versioning

ferenc-csaky:
## Main changes
- Added `scripts/api` to `package.json` and adapted the JSON schema. Now if `api` is present, `graphql` or `operations` cannot be added on their own.
- Parse this data from `package.json` properly
- Refactor the relevant engines to handle multiple schemas
- Work with versioned GraphQL entries in `DatasqrlTest`


sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/config/ServletConfig.java
```diff
@@ -62,4 +62,24 @@ public ServletConfig(JsonObject json) {
...
+
+  public String getRestEndpoint(String version) {
+    return getVersioned(version, restEndpoint);
+  }
+
+  public String getMcpEndpoint(String version) {
+    return getVersioned(version, mcpEndpoint);
+  }
+
+  private String getVersioned(String version, String endpoint) {
```
mbroecheler: make static and reuse in swaggerconfig?

mbroecheler (approved):
Great work. Just one minor comment.

PR #1512, opened 2025-08-26
Canonicalize refactor

ferenc-csaky:
## Main changes

- Path canonicalization removed in the file preprocess phase
- `NameCanonicalizer.SYSTEM` now points to `NameCanonicalizer.AS_IS` instead of `NameCanonicalizer.LOWERCASE_ENGLISH`
- `FlexibleSchemaInferencePreprocessor` now compares a lowercased table/schema name, so the `jsonl` or `csv` does not have to follow the same case pattern than the `.table.sql` and `.schema.yml`
- `OperatorRuleTransformer` still keeps checking for lowercased names to keep function names case insensitive
- `FlinkUdfNsObject`s now keep the original `UpperCamelCase` Java class name instead of `camelCase`-ing them
- Removed some dead code
- Adapted snapshots and test projects


mbroecheler (approved):
Great work. Let's add this to the list of things we need to refactor in datasqrl-examples. It should hopefully be less work since we are also putting all sources into one file to import.

PR #1514, opened 2025-08-28
Extend `enabled-engines` if necessary in `test` cmd instead of replacing them

ferenc-csaky:
## Extending logic

- Always add `test` engine
- Always remove `snowflake` engine
- Add `postgres` if no real DB engine is present
- Make sure `duckdb` is present if `iceberg` is enabled
- Make sure `vertx` is present

Added `ExecutionEngineHolder` to calculate the final exec engine config (most logic from `PipelineFactory`), and now it also does the adapt for `test` exec if needed.


sqrl-planner/src/main/java/com/datasqrl/util/EngineUtil.java
```diff
@@ -0,0 +1,53 @@
...
+  public static List<String> getAvailableQueryEngineNames() {
+    return getChildEngineFactories(QueryEngine.class).map(EngineFactory::getEngineName).toList();
+  }
+
+  public static Stream<EngineFactory> getChildEngineFactories(
+      Class<? extends IExecutionEngine> parentEngineClass) {
+
+    return ServiceLoaderDiscovery.getAll(EngineFactory.class).stream()
+        .filter(factory -> parentEngineClass.isAssignableFrom(factory.getFactoryClass()))
+        .filter(factory -> !factory.getClass().isInterface())
```
mbroecheler: Why does the service loader return interfaces or abstract classes? Are those annotated with AutoLaoder?

sqrl-cli/src/main/java/com/datasqrl/cli/AbstractCompileCmd.java
```diff
@@ -115,7 +115,7 @@ protected void runInternal(ErrorCollector errors) throws Exception {
     }
   }
 
-  protected List<String> getEngines() {
+  protected List<String> getEngines(List<String> enabledEngines) {
     return List.of();
```
mbroecheler: I find this structure confusing (which is not your fault since it was implemented this way already) since it suggests we are getting the engines but what we are actually doing is providing a hook where subclasses can overwrite the engines.

My recommendation is to rename the method to `protected void overwriteEngines` and pass it the sqrlConfig. The base implementation is a no-op. The TestCmd implementation looks at the enabled engines and sets the engines it wants if needed.

sqrl-cli/src/main/java/com/datasqrl/cli/TestCmd.java
```diff
@@ -54,14 +61,24 @@ public ExecutionGoal getGoal() {
   }
 
   @Override
-  protected List<String> getEngines() {
-    return List.of(
-        EngineIds.TEST,
```
mbroecheler, line 59: Thanks for getting rid of EngineIds.

sqrl-cli/src/main/java/com/datasqrl/cli/AbstractCompileCmd.java
mbroecheler: Actually, you can ignore this. See the suggested refactoring below.

sqrl-cli/src/main/java/com/datasqrl/cli/TestCmd.java
```diff
@@ -54,14 +61,24 @@ public ExecutionGoal getGoal() {
...
-        EngineIds.DATABASE,
-        EngineIds.LOG,
-        EngineIds.ICEBERG,
-        EngineIds.DUCKDB,
-        EngineIds.SERVER,
-        EngineIds.STREAMS);
+  protected List<String> getEngines(List<String> enabledEngines) {
+    var engineSet = new HashSet<>(enabledEngines);
```
mbroecheler: On second thought, manipulating engines at the string configuration level isn't great, because it requires a lot of hard-coding of things (not as bad as with the EngineIds, but still).

My recommendation is to handle this inside PipelineFactory. We already do this with defaultEngines there.
inject ExecutionGoal to PipelineFactory and add this to getEngines():
if goal==test, then
a) check if one of the engines is a database engine, if not, add postgres
b) check if one of the engines is a server engine, if not, add vertx

If the user has snowflake configured, they need to create a separate package.json for testing with duckdb. Automatically replacing the engine can be confusing. Instead, we should produce a clear error saying that we cannot run tests against snowflake and they should configure duckdb for testing.

See if we can get away with not having to add the test engine. I could not find where that is being used. PipelineFactory ignores it. It should just be triggered by injected ExecutionGoal imho. That is better than adding a non-engine and we can remove the filter in PipelineFactory and related code.

mbroecheler (changes requested):
brainstorming how we can make this less complex

sqrl-planner/src/main/java/com/datasqrl/util/EngineUtil.java
ferenc-csaky: It does not, I just wrote it pretty much blindly based on the existing classes and ignored the fact how service loader works. :) But you're right. And this logic now is not used at all.

sqrl-cli/src/main/java/com/datasqrl/cli/AbstractCompileCmd.java
ferenc-csaky: I'm a bit confused about a). To make all the already existing usecase pass `postgres` must be present even if `kafka` or `iceberg` are there. Otherwise either the snapshots fail, or the planner.

sqrl-cli/src/main/java/com/datasqrl/cli/TestCmd.java
ferenc-csaky: I'm a bit confused about a). To make all the already existing usecase pass `postgres` must be present even if `kafka` or `iceberg` are there. Otherwise either the snapshots fail, or the planner.

ferenc-csaky: Also, by the time we get to `PipelineFactory`, the compiled `build/package.json` is already written out, and IMO if the content of that does not match with what actually happens that's even worse than overwriting the whole thing as it was before.

sqrl-testing/sqrl-testing-integration/src/test/resources/snapshots/com/datasqrl/UseCaseCompileTest/duckdb--package.txt
```diff
@@ -323,11 +323,11 @@ FROM "_MyApplications2"
             ],
             "query" : {
               "type" : "SqlQuery",
-              "sql" : "SELECT \"id\", 'hello world3' AS \"hello\"\nFROM \"_MyApplications2\"",
+              "sql" : "SELECT \"id\", 'hello world2' AS \"hello\"\nFROM \"iceberg_scan\"('/tmp/duckdb/default_database/_MyApplications2', ALLOW_MOVED_PATHS = TRUE)",
```
mbroecheler: Why is the duckdb test case using postgres now?

ferenc-csaky: The content is the same, just the query order changed. But now i enforced the  to keep the `TreeMap` order in `ExecutionEngineHolder`, so this file also remained the same.

PR #1515, opened 2025-08-28
Fix overwrite column and resolution

mbroecheler:
fix two issues:
1) object identifier resolution did not account for database causing parent tables to not be found
2) AddColumnStatements could not overwrite existing columns. This requires explicitly enumerating all but the added column.
Removed some dead code.


sqrl-planner/src/main/java/com/datasqrl/planner/SqlScriptPlanner.java
```diff
@@ -1075,5 +1084,13 @@ ScriptContext fromImport(ModuleLoader moduleLoader, boolean isInline, String dat
     boolean hasDifferentDatabase(ScriptContext other) {
       return !databaseName.equals(other.databaseName);
     }
+
+    public ObjectIdentifier toIdentifier(String name) {
+      return ObjectIdentifier.of("default_catalog", this.databaseName, name);
+    }
+
+    public ObjectIdentifier toIdentifier(Name name) {
```
ferenc-csaky, line 1092: nit: `ScriptContext` is `private`, any methods there should be package-private

ferenc-csaky (approved):
LGTM, added 1 nitty comment

PR #1517, opened 2025-08-29
Escape SQRL Statements

mbroecheler:
to allow usage of reserved keywords in the sqrl syntactic sugar syntax.


ferenc-csaky (approved):
this makes a lot of sense 👍

PR #1518, opened 2025-08-29
Fix Vert.x service termination on startup failure

velo:
## Summary
- Fixed GraphQLServerVerticle to properly handle startup failures and fail promise correctly
- Removed Promise parameter from SubscriptionConfigurationImpl to prevent double failure completion
- Added proper exception handling to ensure JVM terminates when verticle deployment fails
- Updated HttpServerVerticle to call System.exit(1) when GraphQL verticle deployment fails

## Test plan
- [x] Verify compilation of modified modules
- [x] Run unit tests for sqrl-server-vertx-base module
- [x] Run WriteIT integration test to verify changes work correctly
- [x] Code formatted with pre-commit hooks

🤖 Generated with [Claude Code](https://claude.ai/code)


ferenc-csaky (commented):
Thanks, I think this is a great fix! Can we introduce some vertx lifecycle test(s), WDYT?

PR #1526, opened 2025-09-01
Support compiling Java UDFs on the fly with JBang

### LINKED ISSUE #1267: Remove Output from command line
When running the compiler, we don't want to show this anymore:
```
Compiling...this takes about 10 seconds
Picked up JAVA_TOOL_OPTIONS: --add-exports=java.base/sun.net.util=ALL-UNNAMED   --add-exports=java.rmi/sun.rmi.registry=ALL-UNNAMED   --add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED   --add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED   --add-exports=jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED   --add-exports=jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED   --add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED   --add-exports=java.security.jgss/sun.security.krb5=ALL-UNNAMED   --add-opens=java.base/java.lang=ALL-UNNAMED   --add-opens=java.base/java.net=ALL-UNNAMED   --add-opens=java.base/java.io=ALL-UNNAMED   --add-opens=java.base/java.nio=ALL-UNNAMED   --add-opens=java.base/sun.nio.ch=ALL-UNNAMED   --add-opens=java.base/java.lang.reflect=ALL-UNNAMED   --add-opens=java.base/java.text=ALL-UNNAMED   --add-opens=java.base/java.time=ALL-UNNAMED   --add-opens=java.base/java.util=ALL-UNNAMED   --add-opens=java.base/java.util.concurrent=ALL-UNNAMED   --add-opens=java.base/java.util.concurrent.atomic=ALL-UNNAMED   --add-opens=java.base/java.util.concurrent.locks=ALL-UNNAMED
```

When running run/test command, we want to only show warnings and error log messages and see which of those we can eliminate.

ferenc-csaky:
## Main changes

- Install JBang to the `cmd` Docker image
- Add `JBangRunner` to interact with the native `jbang` binary
- Extract the common JAR processing logic to `UdfManifestPreprocessor`
- Add `JBangPreprocessor` to create JAR files from `.java` files that contain `//DEPS` phrases
- Add container test that tests a project that contains 2 JBang UDFs

## Follow-up changes

- Add support for Kotlin scripts as well

---
Fixes #1267 as well

### DIFF
--- a/sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
+++ b/sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
@@ -72,7 +72,7 @@ public class JBangPreprocessor extends UdfManifestPreprocessor {
     }
 
     try (var lines = Files.lines(file)) {
-      return lines.noneMatch(l -> l.startsWith(DEPS_EXPR));
+      return lines.map(String::trim).noneMatch(l -> l.startsWith(DEPS_EXPR));
     }
   }
 


sqrl-cli/Dockerfile.base
```diff
@@ -19,6 +18,7 @@ RUN apt-get update \
             postgresql-${POSTGRES_VERSION}-pgvector \
     && curl -1sLf 'https://dl.redpanda.com/nzc4ZYQK3WRGd9sy/redpanda/cfg/setup/bash.deb.sh' | bash \
     && apt-get install -y --no-install-recommends redpanda \
+    && curl -Ls https://sh.jbang.dev | bash -s - app setup \
```
velo, line 23: We probably want to lock a version, so we rebuild the same image if needed and we get a new hash when we bump version

sqrl-testing/sqrl-testing-container/src/test/resources/jbang/usrlib/MyAsyncScalarFunction.java
```diff
@@ -0,0 +1,47 @@
...
+import com.google.auto.service.AutoService;
+import org.apache.flink.table.functions.AsyncScalarFunction;
+import org.apache.flink.table.functions.FunctionContext;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+@AutoService(AsyncScalarFunction.class)
```
velo: If the service is necessary, we could always generate the file on META_INF without the `@AutoService`

sqrl-cli/Dockerfile.base
ferenc-csaky: Good catch, i actually made Claude write that, but seems me or them messed something up and it disappeared. :)

sqrl-testing/sqrl-testing-container/src/test/resources/jbang/usrlib/MyAsyncScalarFunction.java
mbroecheler: This is a good point. Is there a way we could skip the `@Autoservice` for JBang, @ferenc-csaky? The authoritative source of what functions to load is in the .json files that we load from the folder.
For generic jars, we are dependent on META_INF for finding those and hence default to @AutoService to make the implementation easier.
I wonder if there is a simple way that with JBang (and the simplifying assumptions that we make) we can generate those function json files more easily and directly. That would allow us to drop the AutoService dependency and make things quite a bit simpler.

sqrl-cli/Dockerfile.base
```diff
@@ -19,6 +20,7 @@ RUN apt-get update \
...
     && apt-get clean && rm -rf /var/lib/apt/lists/*
```
mbroecheler, line 24: In addition to instaling jbang, do we want to initialize some common maven dependencies (i.e. Flink and others) so we don't need to pull them on every invocation and reduce the build time?
What's been your experience @ferenc-csaky in terms of times? Flink deps are pretty heavy, hence I wonder if we can make this a lot faster for simple functions.

sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
```diff
@@ -0,0 +1,59 @@
...
+
+    } catch (ExecuteException e) {
+      log.warn("JBang export failed with exit code: {} for file: {}", e.getExitValue(), file);
+    } catch (IOException e) {
+      log.warn("Could not execute JBang for file: " + file, e);
+    }
+  }
+
+  private boolean skipFile(String fileName) {
+    return !fileName.endsWith(".java") && !fileName.endsWith(".kt");
```
mbroecheler: We should probably test whether the file contains the `//DEPS` keyword. Otherwise we would try to compile every java file (even for nested maven projects), right?

mbroecheler (commented):
Nice work @ferenc-csaky. Added some items to see if we can simplify this further, but the code and overall approach look solid.

sqrl-testing/sqrl-testing-container/src/test/resources/jbang/usrlib/MyAsyncScalarFunction.java
ferenc-csaky: I added a new commit, where I changed this logic to extract the necessary metadata for the `.function.json` via regex text match and eliminated the `@AutoService` dep. There is a lightweight lib that can compile java, but that means an extra dep and for this specific use-case i think the applied approach is good enough, so we can spare introducing a compiler lib for it.

sqrl-cli/Dockerfile.base
ferenc-csaky: It downloads the `flink-table-common` dep in 3sec, and then it reuses the already downloaded deps. So I think until we see some cumbersome examples this should be sufficient for starters.

sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
```diff
@@ -0,0 +1,121 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.exec.ExecuteException;
+import org.apache.commons.io.FilenameUtils;
+
+@RequiredArgsConstructor(onConstructor_ = @Inject)
+@Slf4j
+public class JBangPreprocessor extends UdfManifestPreprocessor {
+
+  private static final Pattern PACKAGE_PATTERN = Pattern.compile("package\\s+([\\w.]+);");
+  private static final Pattern CLASS_PATTERN = Pattern.compile("public\\s+class\\s+(\\w+)");
```
mbroecheler: Consolidate the CLASS_PATTERN and the EXTENDS_PATTERN into one regex to make it more robust. Right now, the former matches any class definition. Aggregation functions typically define an inner class for the aggregated object which might break this.

sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @SneakyThrows
+  private boolean skipFile(Path file) {
+    if (!file.getFileName().toString().endsWith(".java")) {
+      return true;
+    }
+
+    try (var lines = Files.lines(file)) {
+      return lines.noneMatch(l -> l.startsWith(DEPS_EXPR));
```
mbroecheler: Do lines.trim() in case there is whitespace.

sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
```diff
@@ -0,0 +1,121 @@
...
+
+  @Override
+  public void process(Path file, FilePreprocessingPipeline.Context ctx) {
+    if (!jBangRunner.isJBangAvailable() || skipFile(file)) {
+      return;
+    }
+
+    var content = readFileContent(file);
+    if (!isFlinkUdfClass(content)) {
+      log.debug("Skip preprocessing file, as it is not a valid Flink UDF: {}", file);
```
mbroecheler: Should this be a warning? The user probably intended for this to be a Flink UDF.

sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+
+  @SneakyThrows
+  private String readFileContent(Path file) {
+    return Files.readString(file);
+  }
+
+  private boolean isFlinkUdfClass(String content) {
+    var extendsMatcher = EXTENDS_PATTERN.matcher(content);
```
mbroecheler: Should we log an error and return false if a class contains multiple matches (e.g. they define a Udfs class with multiple public static udf implementations contained therein) as tell them to use @AutoService instead?

mbroecheler (changes requested):
I really like the simplicity of the regex matching. That's great for most use cases. Some suggestions for how to make it more robust.

mbroecheler (approved):
Nice!

PR #1528, opened 2025-09-02
Use Jackson instead of Vert.x `JsonObject` in the server config

### LINKED ISSUE #1483: Refactor Vert.x `JsonObject` interface in the server and use Jackson
Right now we use `JsonObject` a lot to put together the final `ServerConfig` instance. It is a horrible interface that adds a lot of boilerplate and we should get away from it.

We should prioritize Jackson everywhere.

### LINKED ISSUE #1484: Generalize query engine URL mapping for Vert.x
Currently, DuckDB and Snowflake are specifically baked in when we init the Vert.x server, but once we get away from `JsonObject`, this should be generalized to be able to fetch any kind of JDBC related data a for all enabled query engines in the pipeline.

ferenc-csaky:
## Main changes

- Remove the `JsonObject` boilerplate from our custom `ServerConfig` classes and use Jackson to ser/deser instead
- Add `QueryEngineConfigConverter` to generalize query engine config conversion between `package.json` and the `server-config.json` format

Fixes #1483
Fixes #1484


ferenc-csaky:
FYI, all tests are succeeded, the tasks are failing because of codecov

PR #1532, opened 2025-09-03
Convert most Lombok `@Value` classes to records, remove some dead code

ferenc-csaky:


velo:
this is awesome!

sqrl-discovery/src/main/java/com/datasqrl/discovery/stats/FieldTypeStats.java
```diff
@@ -53,22 +53,21 @@ public FieldTypeStats(FieldTypeStats other) {
   }
```
velo, line 53: I think this could be a record too.

sqrl-planner/src/main/java/com/datasqrl/config/RootPath.java
```diff
@@ -18,12 +18,8 @@
...
-import lombok.Getter;
 
-@Getter
-public class RootPath {
-
-  private final Path rootDir;
+public record RootPath(Path rootDir) {
 
   @Inject
   public RootPath(@Named("rootDir") Path rootDir) {
```
velo, line 25: wouldn't be better to annotated the field on record with `@Named("rootDir")` 
and use `@AllArgConstructor` to generate constructo with `@Inject`

DI with java records look odd

velo (approved):

sqrl-planner/src/main/java/com/datasqrl/config/RootPath.java
ferenc-csaky: Lombok ctor annotations cannot be put on records.

sqrl-discovery/src/main/java/com/datasqrl/discovery/stats/FieldTypeStats.java
ferenc-csaky: It actually has mutable fields, which are utilized too.

PR #1539, opened 2025-09-05
Fix setting versioned endpoints to `GraphiQLHandlerOptions`

### LINKED ISSUE #1538: GraphiQL URLs broken
Due to the added versioning in 0.8.0. Also check swagger URLs.

ferenc-csaky:
Fixes #1538

### DIFF
--- a/sqrl-testing/sqrl-testing-container/src/test/java/com/datasqrl/container/testing/VertxContainerIT.java
+++ b/sqrl-testing/sqrl-testing-container/src/test/java/com/datasqrl/container/testing/VertxContainerIT.java
@@ -90,9 +90,9 @@ public class VertxContainerIT extends SqrlContainerTestBase {
         .contains("<title>GraphiQL</title>")
         .contains("window.VERTX_GRAPHIQL_CONFIG")
         .contains("\"httpEnabled\":true")
-        .contains("\"graphQLUri\":\"/graphql\"")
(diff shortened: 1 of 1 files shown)


sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/config/ServerConfigUtil.java
```diff
@@ -57,4 +59,41 @@ public static ServerConfig mergeConfigs(
...
+  /**
+   * Creates a copy of GraphiQL handler options with versioned URIs.
+   *
+   * @param version the version prefix to prepend to endpoints
+   * @param graphiQLHandlerOptions the original options to copy and version
+   * @return a new GraphiQL handler options instance with versioned endpoints, or null if input is
+   *     null
+   */
+  @Nullable
+  public static GraphiQLHandlerOptions createVersionedGraphiQLHandlerOptions(
```
mbroecheler, line 72: Do we need to do something similar with Swagger?

ferenc-csaky: nope, the `SwaggerConfig` class is completely our own POJO, so we do not need to hack around built-in vertx things.

PR #1551, opened 2025-09-09
Add config option to tune Flink predicate pushdown related rules

ferenc-csaky:
## Key Changes
- Add `predicate-pushdown-rules` compiler config option with 3 possible values (`DEFAULT`, `LIMITED_TABLE_SOURCE_RULES`, `LIMITED_RULES`)
- The behavior of each possible value is documented in the added `PredicatePushdownRules` enum
- For now the default value is `LIMITED_TABLE_SOURCE_RULES`, which means all types of predicate pushdown (filter, partition, projection) that can hurt subgraph elimination, are disabled for any table source

## TODO

- [x] Add more test cases to `PredicatePushdownTest`

### DIFF
--- a/sqrl-testing/sqrl-testing-integration/src/test/java/com/datasqrl/PredicatePushdownTest.java
+++ b/sqrl-testing/sqrl-testing-integration/src/test/java/com/datasqrl/PredicatePushdownTest.java
@@ -73,7 +73,7 @@ class PredicatePushdownTest extends AbstractAssetSnapshotTest {
   @ParameterizedTest
   @EnumSource(PredicatePushdownRules.class)
   void testSingleFile(PredicatePushdownRules rules) throws Exception {
-    var sqlFile = SCRIPT_DIR.resolve("filter-set-op-transpose.sql");
+    var sqlFile = SCRIPT_DIR.resolve("multi-table-source.sql");
     assertThat(sqlFile).isRegularFile();
 
     executeTest(sqlFile, rules);
@@ -92,7 +92,7 @@ class PredicatePushdownTest extends AbstractAssetSnapshotTest {
       if (it.hasNext()) {
         tEnv.executeSql(stmt.get());
       } else {
-        // Explain the last SELECT
+        // Explain the last SELECT to get the physical plan
         planText = tEnv.explainSql(stmt.get(), ExplainFormat.TEXT);
       }
     }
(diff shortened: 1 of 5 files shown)


sqrl-planner/src/main/java/com/datasqrl/planner/FlinkPlannerConfigBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.flink.table.planner.plan.rules.logical.FlinkFilterJoinRule;
+import org.apache.flink.table.planner.plan.rules.logical.FlinkFilterProjectTransposeRule;
+
+@RequiredArgsConstructor
+public class FlinkPlannerConfigBuilder {
+
+  /** Rules to remove from the logical program to disable filter pushdown. */
+  private static final List<Class<? extends RelOptRule>> FILTER_RULES_TO_REMOVE =
+      List.of(
+          FlinkFilterJoinRule.FILTER_INTO_JOIN.getClass(),
```
mbroecheler: Can we keep that one in there? This is a critical rule for join optimization: without it conditions in the WHERE clause don't get pushed into the JOIN. If you remove this, does it not work anymore?

mbroecheler: I'm singling out this one because people commonly write the join condition in a where clause.

sqrl-planner/src/main/resources/default-package.json
```diff
@@ -3,8 +3,9 @@
   "enabled-engines": ["vertx", "postgres", "kafka", "flink"],
   "compiler": {
     "logger": "print",
-    "extended-scalar-types": true,
     "compile-flink-plan": true,
+    "extended-scalar-types": true,
+    "predicate-pushdown-rules": "DEFAULT",
```
mbroecheler: Let's default this to "LIMITED". It's such a PITA in streaming mode ;-)

Please add documentation for this config option and point out that it does NOT apply in batch mode.

sqrl-testing/sqrl-testing-integration/src/test/resources/predicate-pushdown/table-source-projection-pushdown.sql
```diff
@@ -0,0 +1,14 @@
...
+    v INT,
+    dt STRING
+) PARTITIONED BY (dt)
+WITH (
+    'connector' = 'filesystem',
+    'path' = 'file://__tempdir__',
+    'format' = 'csv',
+    'source.monitor-interval' = '1s'
+);
+
```
mbroecheler, line 13: Could you please make this test case a little more complex by adding something along those lines:

```
CREATE VIEW MyView AS SELECT CONCAT(country,'-x') AS mycountry, country, v+5 AS z, dt, a FROM fs_t;

SELECT mycountry AS country, z FROM MyView
UNION ALL
SELECT dt AS country, a AS z FROM MyView;

PR #1553, opened 2025-09-09
Use awaitility in `DatasqrlTest` to wait/poll Flink job

ferenc-csaky:


sqrl-cli/src/main/java/com/datasqrl/cli/DatasqrlTest.java
```diff
@@ -162,28 +163,26 @@ public int run() {
...
-            } catch (Exception e) {
-              break;
-            }
-
-            Thread.sleep(1000);
-          }
-
+          await()
+              .atMost(delaySec, TimeUnit.SECONDS)
+              .pollInterval(1, TimeUnit.SECONDS)
```
velo, line 168: I would reduce this pool interval, since now is so easy to change it.

PR #1554, opened 2025-09-10
Map mutation schemas in `SwaggerService`

ferenc-csaky:
I tested this manually.

I do not want to hold 0.8.1 further, and include mutation testing to `SwaggerContainerIT` will require a bigger refactor, cause the current setup does not support Kafka on the server img, hence it cannot start up properly.


velo (approved):

ferenc-csaky:
Added https://github.com/DataSQRL/sqrl/issues/1555 to be able to add an integration test for this.

PR #1583, opened 2025-09-23
Fix iceberg database property name in `DuckDbStatementFactory`

ferenc-csaky:
Now, if the table sits in the non-default DB, all of our DuckDB queries will fail.

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/DuckDbStatementFactory.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/DuckDbStatementFactory.java
@@ -25,6 +25,7 @@ import com.datasqrl.calcite.dialect.ExtendedPostgresSqlDialect;
 import com.datasqrl.calcite.type.TypeFactory;
 import com.datasqrl.config.JdbcDialect;
 import com.datasqrl.config.PackageJson.EngineConfig;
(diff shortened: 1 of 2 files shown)


sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/DuckDbStatementFactory.java
```diff
@@ -84,30 +87,30 @@ public QueryResult createQuery(
...
-                        ? "default_database"
-                        : connector.get("database-name");
-                RexBuilder rexBuilder = new RexBuilder(new TypeFactory());
+                var tableId = scan.getTable().getQualifiedName().get(2);
+                var createTable = tableIdMap.get(tableId);
+                var connector = createTable.table().getConnectorOptions();
+
+                var warehouse = connector.get(WAREHOUSE_KEY);
+                var databaseName =
+                    connector.getOrDefault(DATABASE_KEY, SqrlConstants.FLINK_DEFAULT_DATABASE);
```
velo, line 96: The only change, seems to be before we had
```
connector.get("database-name")
```

now we have:
```
connector.get("catalog-database")
```

ferenc-csaky: correct, the rest is just some non-logic affecting refactor

PR #1607, opened 2025-09-29
Map GraphQL list query arguments to array on Vert.x

### LINKED ISSUE #1603: Fix `ARRAY<STRING>` cast for GraphQL query arguments
Currently, when we have a SQRL function that defines an `ARRAY<STRING>` param, if we use GraphQL API to query the endpoint based on some values, it fails with a class type exception.

### SQRL script to reproduce
```sql
StringValues := SELECT val FROM (VALUES ('a'), ('b'), ('c')) AS T(val);

ArrayFn(vals ARRAY<STRING> NOT NULL) := SELECT val FROM StringValues WHERE array_contains(CAST(:vals AS ARRAY<STRING>), val);
```

### GraphQL query
```graphql
{
  ArrayFn(vals: ["a"]) {
    val
  }
}
```

### Exception
```json
{
  "errors": [
    {
      "message": "Exception while fetching data (/ArrayFn) : Parameter at position[0] with class = [java.util.ArrayList] and value = [[a]] can not be coerced to the expected class = [[Ljava.lang.String;] for encoding.",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "ArrayFn"
      ],
      "extensions": {
        "classification": "DataFetchingException"
      }
    }
  ],
  "data": {
    "ArrayFn": null
  }
}
```

ferenc-csaky:
I am not sure this would be required for all query engines, for Postgres it is and fixes the `ARRAY` typed GQL queries.

Also, there is no test for this, cause apparently this does not get hit with a `/*+test*/` query, only by executing the query via GraphiQL web UI (or REST I guess).

Fixes #1603


velo (approved):
LGTM, but, I think you should take that example you made and add as integration test

```
StringValues := SELECT val FROM (VALUES ('a'), ('b'), ('c')) AS T(val);

ArrayFn(vals ARRAY<STRING> NOT NULL) := SELECT val FROM StringValues WHERE array_contains(CAST(:vals AS ARRAY<STRING>), val);

/*+test*/
ArrayFnTest := SELECT val FROM TABLE(ArrayFn(ARRAY['a', 'c']));

```


also, wondering if there are any other types we need to address

Do date and Long which are extended graphql types work well?

Complex input objects we already use a lot, so they do work (use sure if there are tests covering it, but works).

ferenc-csaky:
> LGTM, but, I think you should take that example you made and add as integration test
> 
> ```
> StringValues := SELECT val FROM (VALUES ('a'), ('b'), ('c')) AS T(val);
> 
> ArrayFn(vals ARRAY<STRING> NOT NULL) := SELECT val FROM StringValues WHERE array_contains(CAST(:vals AS ARRAY<STRING>), val);
> 
> /*+test*/
> ArrayFnTest := SELECT val FROM TABLE(ArrayFn(ARRAY['a', 'c']));
> ```
> 
> also, wondering if there are any other types we need to address
> 
> Do date and Long which are extended graphql types work well?
> 
> Complex input objects we already use a lot, so they do work (use sure if there are tests covering it, but works).

But integration tests do not trigger this exception, so not really helpful on its own.

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/VertxQueryExecutionContext.java
```diff
@@ -113,6 +114,20 @@ public CompletableFuture<Object> runQuery(ResolvedSqlQuery resolvedQuery, boolea
     return cf;
   }
 
+  @Override
```
mbroecheler, line 117: Are there other types we need to map here? Can we create a quick test for various types (timestamp, long, double, etc) so we have a unit test for those? Does it already exist?

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/jdbc/AbstractQueryExecutionContext.java
```diff
@@ -18,19 +18,36 @@
...
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 public abstract class AbstractQueryExecutionContext<C extends Context>
     extends StandardExecutionContext<C> implements QueryExecutionContext {
 
-  public AbstractQueryExecutionContext(
+  protected AbstractQueryExecutionContext(
```
mbroecheler, line 32: JdbcQueryExecutionContext also extends this, but I wonder if we still need that class and JdbcContext. Seems we only use VertxContext. Can we get rid of the Jdbc* variants?

ferenc-csaky: Nope it's dead code, I removed it.

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/VertxQueryExecutionContext.java
ferenc-csaky: The main problem is there is no quick test to do that, cause the original problem only gets hit if you query from the graphiql UI. `DatasqrlTest` queries work properly.

ferenc-csaky: Added one array test. Started to add more (int, long, double), but then I was wondering that testing all types would be redundant, cause we have other tests that map basic types, and the main problem here was the container type, not the element type. But let me know if i'm wrong.

PR #1623, opened 2025-10-06
Secrets in deployment configuration

mbroecheler:


mbroecheler:
@ferenc-csaky Good point, these are not actually the secrets themselves but the identifier of a secret in the secrets manager that is loaded by the kubernetes external secrets operator (or whatever other mechanism a user of sqrl might have). The configuration says: "In order to run this job, you need to load the following secrets".

PR #1629, opened 2025-10-09
Support sessions for DuckDB queries

ferenc-csaky:
## Key changes
- Added `config` map for `duckdb` engine in `package.json`
- Introduced `use-disk-cache` config option for now
- DuckDB extensions are installed at JDBC client init, not before query execution
- Added test usecase with disk cache enabled


sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/jdbc/SqlClientWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+      super(sqlClient);
+      this.config = config;
+    }
+
+    public DuckDbClientWrapper withClient(SqlClient sqlClient) {
+      return new DuckDbClientWrapper(sqlClient, config);
+    }
+
+    public String getExtensionInstall() {
+      var extInstallStr = "INSTALL iceberg;";
```
mbroecheler: What about the httpfs extension? Do we need that for s3 access and for cache_httpfs to work?

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/jdbc/VertxJdbcClient.java
```diff
@@ -54,20 +54,25 @@ public ResolvedQuery unpreparedQuery(SqlQuery sqlQuery, Context context) {
 
   public Future<RowSet<Row>> execute(
       DatabaseType database, PreparedQuery<RowSet<Row>> query, Tuple tup) {
-    var sqlClient = clients.get(database);
+    var wrapper = clients.get(database);
 
+    // For DuckDB, load extensions per connection (LOAD is idempotent)
```
mbroecheler: Does that comment still apply? It seems like we load extensions on initialization now, right?

mbroecheler: Ah, I see. We install extension on initialization but lead them for every query? Is that efficient? Is it possible to load them on initialization as well? 

This is what GPT5 hallucinated:
How DuckDB extensions work
	•	INSTALL iceberg; → downloads/registers the extension the first time on that machine/database. Safe to run at startup; it’s a no-op after the first install.
	•	LOAD iceberg; → loads the extension into the current connection/session. Once loaded for that connection, you don’t need to run it again for each query. If you run it repeatedly in the same connection, it’s just unnecessary overhead.

Because you’re using a Vert.x pool (i.e., multiple connections), you want to make sure each connection runs LOAD iceberg; exactly once when it’s created/checked-out the first time.

Practical patterns

1) Single shared connection (simple)
	•	Create one DuckDB JDBC connection at Vert.x startup.
	•	Run:
```
INSTALL iceberg;
LOAD iceberg;
```

mbroecheler: As I understand it, DuckDb supports concurrent queries on a connection.

ferenc-csaky: Both `INSTALL` and `LOAD` gets ignored if those are already, so there's no overhead. But if a single connection would be enough, that can be done to, I was trying to keep the pooling.

The crucial part that's wrong right now is the installed extensions are stored in the filesystem on the DuckDB side, so those are applying to any connections, although loading has to be per-connection level if we keep the pool, but without modifying the connection selection, that may loose the previous extension context, so for now I will optimize for a single `SqlConnection`, casue that woul be the least invasive and complex solution, and seems like a good start we can iterate on later if necessary.

PR #1650, opened 2025-10-21
Update to Flink 2.1

ferenc-csaky:


sqrl-testing/sqrl-testing-container/src/test/resources/jbang/usrlib/MyAsyncScalarFunction.java
```diff
@@ -1,4 +1,4 @@
-//DEPS org.apache.flink:flink-table-common:1.19.3
+//DEPS org.apache.flink:flink-table-common:2.1.0
```
velo, line 1: Is it possible for us to make this something like a managed dependency so user don't need to update it?!

sqrl-testing/sqrl-testing-integration/src/test/resources/usecases/repository/repo.sqrl
```diff
@@ -1,14 +1,14 @@
...
+Submission := SELECT *, IFNULL(`variant`, 'default') AS variant0,
                      concat('https://repository.datasqrl.com/',file) AS repoURL
               FROM _SubmissionRaw;
 
 /*+query_by_any(name) */
 Package := SELECT DISTINCT name FROM Submission;
 
 Package.latest := SELECT * FROM Submission s WHERE this.name = s.name AND s.latest = TRUE ORDER BY s.submissionTime DESC LIMIT 1;
-Package.versions(version String, variant String) := SELECT * FROM  Submission s WHERE this.name = s.name AND (:version IS NULL OR :version=s.version) AND (:variant IS NULL OR :variant=s.variant)
+Package.versions(version String, `variant` String) := SELECT * FROM  Submission s WHERE this.name = s.name AND (:version IS NULL OR :version=s.version) AND (:`variant` IS NULL OR :`variant`=s.`variant`)
```
velo, line 11: it would be really nice if sqrl was able to auto quote reserved words

mbroecheler: Yes, unfortunately this is too deep in Calcite. We will revisit this when we have a ton more engineers ;-)

sqrl-testing/sqrl-testing-container/src/test/resources/jbang/usrlib/MyAsyncScalarFunction.java
mbroecheler: The problem is that the user might want to run the class independently of SQRL (e.g. for unit testing)

mbroecheler: But I agree that it is kinda annoying for the user to have to track the version. Is there something we could do to automate this @ferenc-csaky (in a different PR ofc)

PR #1675, opened 2025-11-05
Support server-side function execution

ferenc-csaky:
## Key changes
- Foundation of `FlinkExecFunction` code copied over from #1336, but moved to `sqrl-server-vertx-base`
- Added `FunctionExecutor` and its impl `FlinkFunctionExecutor`, which is wired into the server context, and gets called when a computed parameter is needed
- Added `RowDataMapper` to map between internal Flink types and Java objects
- Introduced full use-case with preprocess server function
- Enforced alphabetic field order serialization on Flink `LogicalType` obejcts to make `vertx-exec-functions.json` deterministic

## TODO
- [x] Check how much extra size including `flink-table-planner` into the server means
- [x] Execute the function on a separate Vert.x worker thread
- [x] Make sure we return the proper container type after fn exec, like list or single object


ferenc-csaky:
Regarding the `vertx-server.jar` size, unsfortunately with all the transitive garbage `flink-table-planner_2.12` fatten the JAR from 122MB to 344MB. Because we need the scala codegen parts, we cannot escape depending on the table planner JAR, but after some experimenting I was able to shave off more than 100MB, so we now have a 239MB `vertx-server.jar`, but at this point we cannot exclude anything else.

PR #1691, opened 2025-11-17
Fix Vert.x Promise race condition in HttpServerVerticle deployment

velo:
## Problem

The `HttpServerVerticle` had a race condition where it would complete its `startPromise` when the HTTP server started listening, but **before** the child `GraphQLServerVerticle` finished deploying. This caused an `IllegalStateException: Promise already completed` error when the GraphQL verticle tried to complete its own promise after subscriptions were ready.

**Error symptoms:**
- `java.lang.IllegalStateException: Promise already completed`
- `sqrl.vertx.core.VertxException: Verticle un-deployed`
- GraphQL verticle deployment failures

## Root Cause

In the `bootstrap()` method:
1. GraphQL verticles were deployed asynchronously via `deployVersionedModel()`
2. HTTP server immediately started and completed `startPromise` 
3. Later, when GraphQL verticle finished setup, it tried to complete the same promise → ERROR

## Solution

Modified the deployment flow to ensure proper ordering:

1. Changed `deployVersionedModel()` to return `Future<String>` instead of `void`
2. Collected all GraphQL verticle deployment futures into a list
3. Used `Future.all()` to wait for all verticles to complete deployment
4. Only start HTTP server **after** all child verticles are successfully deployed

This ensures the parent's `startPromise` is only completed after all child deployments finish.

## Testing

- Built and tested with `datasqrl/cmd:local` image
- Verified Promise errors are completely eliminated
- HTTP server starts correctly after GraphQL verticle deployment completes
- No regression in verticle deployment flow

## Files Changed

- `sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/HttpServerVerticle.java`

### DIFF
--- a/sqrl-testing/sqrl-testing-container/src/test/java/com/datasqrl/container/testing/VertxContainerIT.java
+++ b/sqrl-testing/sqrl-testing-container/src/test/java/com/datasqrl/container/testing/VertxContainerIT.java
@@ -73,7 +73,8 @@ public class VertxContainerIT extends SqrlContainerTestBase {
     var randomRequest = new HttpPost(getBaseUrl() + randomPath);
     randomRequest.setEntity(new StringEntity(testBody, ContentType.APPLICATION_JSON));
 
-    sharedHttpClient.execute(randomRequest);
+    var randomResponse = sharedHttpClient.execute(randomRequest);
+    EntityUtils.consume(randomResponse.getEntity());
 
     assertTraceLogContains("ReqBody:", testBody, randomPath);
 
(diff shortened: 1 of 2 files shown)


sqrl-testing/sqrl-testing-container/src/test/java/com/datasqrl/container/testing/VertxContainerIT.java
```diff
@@ -40,6 +40,7 @@ protected String getTestCaseName() {
   void givenUdfScript_whenCompiledAndServerStarted_thenApiRespondsCorrectly() {
     compileAndStartServer("myudf.sqrl", testDir);
 
+    System.out.println(serverContainer.getLogs());
```
ferenc-csaky: I guess we can get rid of this, it's just bloating STDOUT.

ferenc-csaky (approved):

PR #1695, opened 2025-11-18
Streamline CLI params

### LINKED ISSUE #1692: Streamline CLI to only accept a list of package config files
Currently, the user can either provide a `.sqrl` script, and optionally a GraphQL API definition, or one or multiple package configuration. This can get confusing, so the goal is to simplify the CLI to only take a list of package config files, and the `.sqrl` script, and any API definition has to be set properly in that package config JSON.

ferenc-csaky:
## Key changes
- Merged `PackageBootstrap` into `AbstractCompileCmd`, cause there was 50 lines of code remaining there
- Adapted our automated test framework to be able to work with  `package.json` files exclusively
  - Reworked recursive file collection a bit and put that code into `ArgumentsProviders`
  - `DAGPlannerTest` and `GraphQLValidationTest` has a lot of SQRL scripts and GraphQL schemas in 1 project, to spare creating a bunch of package files for nothing, I added some templating logic -> `AbstractAssetSnapshotTest#writeTempPackage`
- The logic got simpler pretty much everywhere  (We could take this further and make it even more cleaner but I am already happy with the current state for now)

Resolves #1692


sqrl-cli/src/main/java/com/datasqrl/cli/AbstractCompileCmd.java
```diff
@@ -21,50 +21,29 @@
...
-  private static final String GRAPHQL_FILE_NAME_PATTERN = "schema.%s.graphqls";
-
-  @CommandLine.Parameters(
-      arity = "0..2",
-      description = "Main script and (optional) API specification")
-  protected Path[] files = new Path[0];
-
-  @CommandLine.Option(
-      names = {"-a", "--api"},
-      description = "Generates the API specification for the given type")
```
mbroecheler, line 55: I use this flag quite a lot to generate the API schema so that I can fine tune it. It's fine to let the option go and simplify the API, but then we should add the generated artifact to the `build/` directory as `pipeline_api.v1.graphqls`. That way it's always there (and I can copied it to main dir when needed) and it lines up with the other pipeline build artifacts we write out. WDYT?

sqrl-cli/src/main/java/com/datasqrl/cli/AbstractCompileCmd.java
```diff
@@ -126,36 +101,47 @@ protected void postprocess(
...
+  /**
+   * Initializes the {@link PackageJson} configuration by merging all provided package configuration
+   * files. If no files are provided, attempts to load the default package configuration file.
+   *
+   * @param errors error collector
+   * @param rootDir the root project directory
+   * @param packageFiles the list of given package configuration file paths
+   * @return the merged {@link PackageJson} configuration
+   */
+  private PackageJson initPackageJson(
```
mbroecheler, line 113: This used to be in the packager, right? Why did you move it here?

mbroecheler: Ah, I see you got rid of PackageBootstrap - nice

sqrl-cli/src/main/java/com/datasqrl/cli/AbstractCompileCmd.java
ferenc-csaky: This is a good point, and I like the solution you suggested. Since I need to adapt this PR to the current main, cause I merged other test and snapshot related changes, I'll include it into this PR. :)

PR #1700, opened 2025-11-19
Bump `graphql-java` to 25.0 and `graphql-java-extended-scalars` to 24.0

ferenc-csaky:
## Key Changes
- GraphQL introduced strict mode by default, which means you cannot redefine built-in types (what we do for `Float` and `DateTime`), so now we explicitly need to disable that (possible follow-up if this could be achieved differently on our send, so we can roll with strict mode)
- `Coercing` interface method signatures changed, those had to be adapted
- The  data fetching got smarter (and more performant in some scenarios) in recent versions, so now our custom `PropertyDataFetcher` has to override 2 `get` methods
- Also separated our custom property fetcher code into its separate class `CaseInsensitiveJsonDataFetcher`
- `slf4j-api` got removed from `graphql-java` so now we need to add it explicitly into our `sqrl-server-core` to be able to define loggers
- Removed explicit `reactive-streams` dependency definition from our POMs cause it ships with `graphql-java`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -127,8 +127,8 @@
     <feign.version>13.6</feign.version>
     <freemarker.version>2.3.34</freemarker.version>
     <glassfish.version>3.0.0</glassfish.version>
-    <graphql-java-extended-scalars.version>22.0</graphql-java-extended-scalars.version>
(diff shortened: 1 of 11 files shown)


mbroecheler (approved):
Thank you!

PR #1720, opened 2025-11-26
Improve error message in parseSQL when statement count is not 1

velo:
## Summary
- Improved the error message in `Sqrl2FlinkSQLTranslator.parseSQL()` to include helpful debugging information when SQL parsing fails
- Now shows the number of statements found, the SQL being parsed (truncated to 500 chars), and the parsed statements
- Helps diagnose issues like empty SQL strings or multiple statements separated by semicolons

## Test plan
- [x] Unit tests pass (`mvn test -pl sqrl-planner`)
- [x] No new tests needed - this is a debugging improvement

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/planner/Sqrl2FlinkSQLTranslator.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/planner/Sqrl2FlinkSQLTranslator.java
@@ -263,13 +263,13 @@ public class Sqrl2FlinkSQLTranslator {
       throw new RuntimeException(e);
     }
     var sqlNodeList = parser.parseSqlList(sqlStatement);
-    List<SqlNode> parsed = sqlNodeList.getList();
+    var parsed = sqlNodeList.getList();
     checkArgument(
         parsed.size() == 1,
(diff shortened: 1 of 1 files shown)


sqrl-planner/src/main/java/com/datasqrl/planner/Sqrl2FlinkSQLTranslator.java
```diff
@@ -264,7 +264,12 @@ public SqlNode parseSQL(String sqlStatement) {
     }
     var sqlNodeList = parser.parseSqlList(sqlStatement);
     List<SqlNode> parsed = sqlNodeList.getList();
-    checkArgument(parsed.size() == 1);
+    checkArgument(
+        parsed.size() == 1,
+        "Expected exactly 1 SQL statement but found %s. SQL: [%s]. Parsed statements: %s",
+        parsed.size(),
+        sqlStatement.length() > 500 ? sqlStatement.substring(0, 500) + "..." : sqlStatement,
+        parsed);
```
ferenc-csaky: The `parsed` list will pring out the stringified parsed statements again, I think it would be better to include the parsed size only into the message and only print the raw SQL string limited to 500 chars, as it is now.

ferenc-csaky (approved):

PR #1728, opened 2025-12-01
Implement `add-func` CLI command to help creating Java UDFs

ferenc-csaky:
## Key Changes
  - **Implemented `add-func` CLI command**: New command that scaffolds user-defined functions in the `functions` folder with support for both scalar and aggregate function types via `--aggregate` flag.

  - **Added UDF templates**: Created two template files (`scalar.java` and `aggregate.java`) with example implementations that use placeholder substitution for the UDF class name.

  - **Registered command in CLI**: Added `AddFuncCmd` to the `DatasqrlCli` subcommands list between `InitCmd` and `CompileCmd`.

  - **Comprehensive test coverage**: Added 7 unit tests to `AddFuncCmdTest` covering file creation, placeholder replacement, template selection, and content validation.

  - **Template features**: Scalar template includes basic string manipulation example, aggregate template demonstrates time-weighted average with accumulator pattern.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1004,7 +1004,7 @@
               <excludes>
                 <exclude>**/src/test/resources/**</exclude>
                 <exclude>**/resources/templates/init-project/**</exclude>
-                <exclude>**/resources/templates/udfs/**</exclude>
+                <exclude>**/resources/templates/functions/**</exclude>
                 <exclude>**/m2e-target/**</exclude>
                 <exclude>bin/**</exclude>
                 <exclude>documentation/**/index_features_list.txt</exclude>
--- a/sqrl-cli/src/main/java/com/datasqrl/cli/AddFuncCmd.java
+++ b/sqrl-cli/src/main/java/com/datasqrl/cli/AddFuncCmd.java
@@ -30,24 +30,22 @@ import picocli.CommandLine.Option;
 import picocli.CommandLine.Parameters;
 
 @CommandLine.Command(
-    name = "add-udf",
+    name = "add-func",
     description =
-        "Adds a new user-defined function into the 'functions' folder of an existing project")
-public class AddUdfCmd extends BaseCmd {
+        "Adds a new function definition into the 'functions' folder of an existing project")
+public class AddFuncCmd extends BaseCmd {
 
-  private static final String UDF_TEMPLATE_DIR = "templates/udfs";
-  private static final String AGGREGATE_UDF_TEMPLATE = UDF_TEMPLATE_DIR + "/aggregate.java";
-  private static final String SCALAR_UDF_TEMPLATE = UDF_TEMPLATE_DIR + "/scalar.java";
-  private static final String UDF_NAME_PLACEHOLDER = "__udfname__";
+  private static final String FN_TEMPLATE_DIR = "templates/functions";
+  private static final String FN_NAME_PLACEHOLDER = "__fnname__";
+  private static final String AGGREGATE_UDF_TEMPLATE = FN_TEMPLATE_DIR + "/aggregate.java";
+  private static final String SCALAR_UDF_TEMPLATE = FN_TEMPLATE_DIR + "/scalar.java";
 
-  @Parameters(
-      index = "0",
-      description = "Name of the newly added user-defined Java file and function class")
-  String udfName;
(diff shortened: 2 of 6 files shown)


sqrl-cli/src/main/java/com/datasqrl/cli/AddUdfCmd.java
```diff
@@ -0,0 +1,74 @@
...
+import java.nio.file.StandardOpenOption;
+import java.util.function.Supplier;
+import lombok.SneakyThrows;
+import org.apache.commons.io.IOUtils;
+import picocli.CommandLine;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.Parameters;
+
+@CommandLine.Command(
+    name = "add-udf",
```
mbroecheler: How do you feel about just calling it `add-function` it's a bit longer but probably clearer. We throw the term UDF around in Flink but from the context of the user what they are trying to do is implement a function. The fact that it is "user defined" is implicit from the context ;-)

sqrl-cli/src/main/java/com/datasqrl/cli/AddUdfCmd.java
```diff
@@ -0,0 +1,74 @@
...
+public class AddUdfCmd extends BaseCmd {
+
+  private static final String UDF_TEMPLATE_DIR = "templates/udfs";
+  private static final String AGGREGATE_UDF_TEMPLATE = UDF_TEMPLATE_DIR + "/aggregate.java";
+  private static final String SCALAR_UDF_TEMPLATE = UDF_TEMPLATE_DIR + "/scalar.java";
+  private static final String UDF_NAME_PLACEHOLDER = "__udfname__";
+
+  @Parameters(
+      index = "0",
+      description = "Name of the newly added user-defined Java file and function class")
```
mbroecheler: Simplify to "Name of the function"

sqrl-cli/src/main/java/com/datasqrl/cli/AddUdfCmd.java
```diff
@@ -0,0 +1,74 @@
...
+import lombok.SneakyThrows;
+import org.apache.commons.io.IOUtils;
+import picocli.CommandLine;
+import picocli.CommandLine.Option;
+import picocli.CommandLine.Parameters;
+
+@CommandLine.Command(
+    name = "add-udf",
+    description =
+        "Adds a new user-defined function into the 'functions' folder of an existing project")
```
mbroecheler: "Adds a new function definition into the 'functions' folder of an existing project"

sqrl-cli/src/main/java/com/datasqrl/cli/AddUdfCmd.java
```diff
@@ -0,0 +1,74 @@
...
+  private static final String UDF_NAME_PLACEHOLDER = "__udfname__";
+
+  @Parameters(
+      index = "0",
+      description = "Name of the newly added user-defined Java file and function class")
+  String udfName;
+
+  @Option(
+      names = {"--aggregate"},
+      description = "Add an aggregate UDF function instead of a scalar one")
```
mbroecheler: "Adds an aggregation functions instead of the default scalar one"

sqrl-cli/src/main/java/com/datasqrl/cli/AddUdfCmd.java
```diff
@@ -0,0 +1,74 @@
...
+  boolean aggregate = false;
+
+  @Override
+  protected void runInternal(ErrorCollector errors) {
+    try {
+      addUdf(() -> cli.rootDir);
+    } catch (Exception e) {
+      throw errors.exception("Failed to add UDF %s: %s", udfName, e);
```
mbroecheler: Failed to add function:

sqrl-cli/src/main/java/com/datasqrl/cli/AddUdfCmd.java
ferenc-csaky: What about `add-func`? 😄 I think that's a common-enough shortening of function, and we can write it out properly in every description.

mbroecheler: Sounds good. You are sooo lazy ;-)

PR #1733, opened 2025-12-02
Add OAuth 2.0 authentication support for MCP server

velo:
## Summary
- Add OAuth 2.0 Protected Resource Metadata (RFC 9728) support for MCP endpoints
- Enable JWKS-based token validation via OpenID Connect discovery
- Support flexible auth configuration: JWT only, OAuth only, none, or both
- Add `/.well-known/oauth-protected-resource` discovery endpoint
- Return 401 with `WWW-Authenticate` header for OAuth-protected endpoints

## Changes
- New `OAuthConfig` configuration class for OAuth settings
- New `OAuthDiscoveryHandler` for RFC 9728 discovery endpoint
- New `OAuth2AuthFactory` for OIDC-based auth provider creation
- New `OAuthFailureHandler` with WWW-Authenticate header support
- Updated `enabledAuth` array in package.json schema (replaces deprecated `authKind`)
- Added Keycloak-based integration test (`McpOAuthIT`)
- Updated documentation with OAuth configuration examples

## Test plan
- [x] Unit tests pass
- [ ] Integration test (`McpOAuthIT`) requires Docker images to be rebuilt with schema changes
- [x] ServerConfigTemplateTest validates new `oauthConfig` field serialization

### DIFF
--- a/documentation/docs/configuration-engine/vertx.md
+++ b/documentation/docs/configuration-engine/vertx.md
@@ -67,7 +67,7 @@ For example, for `ES256`, this would look something like this:
 
 ## OAuth 2.0 Authentication Configuration
 
-For OAuth 2.0 authentication with providers like Auth0 or Keycloak, use the `oauth2Options` section.
+For OAuth 2.0 authentication with providers like Auth0 or Keycloak, use the `oauthConfig` section.
 This enables MCP (Model Context Protocol) clients like Claude Code to authenticate using OAuth.
 
 ```json
@@ -76,11 +76,11 @@ This enables MCP (Model Context Protocol) clients like Claude Code to authentica
     "vertx": {
       "authKind": ["OAUTH"],
       "config": {
-        "oauth2Options": {
-          "site": "https://your-tenant.auth0.com/",
-          "clientId": "your-client-id"
-        },
-        "oauthDiscovery": {
+        "oauthConfig": {
+          "oauth2Options": {
+            "site": "https://your-tenant.auth0.com/",
+            "clientId": "your-client-id"
+          },
           "authorizationServerUrl": "https://your-tenant.auth0.com/",
           "scopesSupported": ["mcp:tools", "mcp:resources"]
         }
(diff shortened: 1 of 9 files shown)


documentation/docs/configuration-engine/vertx.md
```diff
@@ -4,18 +4,19 @@ Eclipse Vert.x is a reactive server framework that serves as the GraphQL API ser
 
 ## Configuration Options
 
-| Key        | Type       | Default   | Notes                                                |
-|------------|------------|-----------|------------------------------------------------------|
-| `authKind` | **string** | `"NONE"`  | Authentication type: `"NONE"` or `"JWT"`             |
-| `config`   | **object** | see below | Vert.x-specific configuration including JWT settings |
+| Key           | Type        | Default   | Notes                                                         |
+|---------------|-------------|-----------|---------------------------------------------------------------|
+| `enabledAuth` | **array**   | `[]`      | List of enabled auth methods: `"JWT"`, `"OAUTH"`              |
```
ferenc-csaky: I do not really like `enabledAuth` as a name. IMO it is confusing, cause that name implicates a boolean flag. I'd rather use `auth`, `authMethods`, or just update `authKind`.

Whatever we do, I think we should not keep the deprecated field, 0.9 can still contain breaking changes.

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/config/OAuthConfig.java
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * OAuth 2.0 configuration for MCP server authentication. Supports Auth0, Keycloak, and other OAuth
+ * 2.0 / OpenID Connect providers.
+ */
+@Getter
+@Setter
+@NoArgsConstructor
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class OAuthConfig {
```
ferenc-csaky, line 37: I don't really understand why is this whole class necessary. Vertx provides `OAuth2Options`, so we should simply use that in `ServerConfig`, just as we do with `JWTAuthOptions`.

plan/mcp-oauth.md
ferenc-csaky, line 1: Can you make this doc file up to date too? And what's the purpose of this? Yet another file yet another place, nobody will find this if it is intended for internal use :D

documentation/docs/configuration-engine/vertx.md
```diff
@@ -4,18 +4,18 @@ Eclipse Vert.x is a reactive server framework that serves as the GraphQL API ser
...
+| `config`   | **object**| see below| Vert.x-specific configuration including auth settings   |
 
 ## Basic Configuration
 
 ```json
 {
   "engines": {
     "vertx": {
-      "authKind": "NONE"
+      "authKind": []
```
ferenc-csaky, line 18: What happens if the `authKind` is `null`? I think we should produce the same behavior as if it's an empty array, is it how it goes now?

PR #1734, opened 2025-12-03
Surface captured errors when tests fail (Maven-like approach)

velo:
## Summary
- Added `ErrorCapturingStream` class that captures exception and error lines from stdout/stderr during test execution
- Modified `TestOutputManager` to use error capturing and expose captured errors
- When tests fail, errors are now displayed in a "Captured Errors" section before the "Test Reports" section

This addresses the issue where exceptions during test execution (like the Flink `RowTime field should not be null` error) were only visible in the log file, making debugging difficult. Now they're surfaced directly in the console output when tests fail, similar to how Maven shows errors.

## Test plan
- [x] Added unit tests for `ErrorCapturingStream` covering:
  - Normal output (no errors captured)
  - RuntimeException capture with stack traces
  - Caused by exception capture
  - Error class capture
  - Clear functionality
  - Stack trace truncation for long traces
- [x] Verified code compiles successfully
- [x] Pre-commit hooks passed (code formatting)


sqrl-cli/src/main/java/com/datasqrl/cli/TestCmd.java
```diff
@@ -53,16 +57,28 @@ protected void execute(ErrorCollector errors) throws Exception {
...
+            "Test snapshots:     " + sqrlConfig.getTestConfig().getSnapshotDir(cli.rootDir));
+        formatter.info("");
+      }
+
       formatter.buildStatus(success, getElapsedTime(), LocalDateTime.now());
 
       if (!success) {
-        formatter.helpLink("Help 1", "https://docs.datasqrl.com/docs/howto/testing");
+        formatter.helpText("Run with --update-snapshots to update expected results.");
+        formatter.helpLink("Help 1", "https://datasqrl.com/docs/testing#snapshot-failures");
```
ferenc-csaky: This is both wrong, so teach Claude that. `--update-snapshots` is not something the CLI support, and that link is wrong too. I fixed this yesterday, keep that pls.

sqrl-cli/src/main/java/com/datasqrl/cli/TestCmd.java
```diff
@@ -35,9 +36,12 @@ protected void execute(ErrorCollector errors) throws Exception {
     }
 
     try (var outputMgr = new TestOutputManager(cli.rootDir)) {
+      outputMgr.init();
       outputMgr.disableConsoleLogs();
 
-      var formatter = getOutputFormatter();
+      var formatter =
+          new DefaultOutputFormatter(
+              batchMode, outputMgr.getOriginalOut(), outputMgr.getOriginalErr());
```
ferenc-csaky: This should be `getOutputFormatter()`. And `getOriginalOut()` and `getOriginalErr()` are meaningless, they will return `System.out` and `System.err`, which the 1 param ctor will do, which `getOutputFormatter()` will return. But skipping the maven styled format should be possible, cause it just makes more mess for the internal test runs, and I don't like it in that specific case.

sqrl-testing/sqrl-testing-integration/src/test/java/com/datasqrl/AbstractFullUseCaseTest.java
```diff
@@ -133,10 +132,8 @@ void fullUseCaseTest(UseCaseParam param) {
               .resolve(SqrlConstants.DEPLOY_DIR_NAME)
               .resolve(SqrlConstants.PLAN_DIR);
       var flinkConfig = loadInternalTestFlinkConfig(planDir, env);
-      var outputMgr = new TestOutputManager(rootDir);
-      var test =
-          new DatasqrlTest(
-              rootDir, planDir, packageJson, flinkConfig, env, outputMgr, new NoOutputFormatter());
+      var formatter = new DefaultOutputFormatter(true);
+      var test = new DatasqrlTest(rootDir, planDir, packageJson, flinkConfig, env, formatter);
```
ferenc-csaky: I'd keep `NoOutputFormatter()` here, there are specific IT cases to test the output format, and I don't like the maven formatting in our CI, which is maven as well, and that makes the whole thing more complicated to read.

sqrl-cli/src/main/java/com/datasqrl/cli/output/TestOutputManager.java
```diff
@@ -31,30 +32,31 @@ public class TestOutputManager implements AutoCloseable {
...
+  private final PrintStream originalOut = System.out;
+  private final PrintStream originalErr = System.err;
   private PrintStream logStream;
+  private ErrorCapturingStream errorCapturingStream;
 
-  public void redirectStd() {
-    originalOut = System.out;
-    originalErr = System.err;
+  @SneakyThrows
+  public void init() {
```
ferenc-csaky: I don't see any reason for this `init()`. We instantiate this object in `TestCmd` and call `init()` right after, so lets just move the `init()` logic to the ctor. Then `getCapturedErrors` can be simplified to a lombok getter as well.

sqrl-cli/src/main/java/com/datasqrl/cli/output/TestOutputManager.java
```diff
@@ -31,30 +32,31 @@ public class TestOutputManager implements AutoCloseable {
...
+    return originalOut;
+  }
 
-    if (originalErr != null) {
-      System.setErr(originalErr);
-      originalErr = null;
-    }
+  public PrintStream getOriginalErr() {
+    return originalErr;
   }
```
ferenc-csaky: These getter are unnecessary, see my other comment about it on the `TestCmd` file.

sqrl-cli/src/main/java/com/datasqrl/cli/output/ErrorCapturingStream.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * An OutputStream that writes to a delegate stream while capturing lines that appear to be errors
+ * or exceptions. This allows surfacing important error messages to users when tests fail, similar
+ * to how Maven shows errors at the end of a failed build.
+ */
+public class ErrorCapturingStream extends OutputStream {
+
+  private static final int MAX_CAPTURED_LINES = 100;
+  private static final int MAX_STACK_TRACE_LINES = 20;
```
ferenc-csaky: I'm not sure we should limit these. A half error a truncated stack trace can still make no sense. So I'd rather not limit any of them. I think if these lines get too big we have bigger problems anyways. :)

sqrl-cli/src/main/java/com/datasqrl/cli/TestCmd.java
velo: ow, this was me, I screw up when I rebased this branch

sqrl-cli/src/main/java/com/datasqrl/cli/DatasqrlTest.java
```diff
@@ -268,6 +272,14 @@ private void printTestResults(List<TestResult> testResults, String snapshotDir)
...
         }
       }
+
+      outputManager.printCapturedErrors(formatter);
+
+      formatter.sectionHeader("Test Reports");
+      formatter.info("Full execution log: build/logs/test-execution.log");
+      formatter.info("Flink metrics:      build/logs/flink-metrics.log");
+      formatter.info("Test snapshots:     " + snapshotDir);
+      formatter.info("");
```
ferenc-csaky, line 282: nit: `formatter.newline()`

PR #1742, opened 2025-12-04
Support DuckDB queries from S3

ferenc-csaky:
## Key Changes
- In case of Glue, the S3 bucket path contains a `.db` suffix for the database name, which now `DuckDbStatementFactory` handles properly
- Added case of `use-version-guessing` config options where we call `SET unsafe_enable_version_guessing = true` during init, cause Glue does not work with `scan_iceberg(...)` calls otherwise
- Started to use `HikariCP` as data source when we init the DuckDB JDBC pool to properly execute session-wide preconfigurations
- Reworked config keys and organization
- Added docs

### DIFF
--- a/sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/util/DuckDbExtensions.java
+++ b/sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/util/DuckDbExtensions.java
@@ -15,7 +15,7 @@
  */
 package com.datasqrl.util;
 
-import java.util.Map;
+import com.datasqrl.graphql.config.JdbcConfig;
 import java.util.StringJoiner;
 import lombok.RequiredArgsConstructor;
 
@@ -24,7 +24,7 @@ public final class DuckDbExtensions {
 
   private final StringJoiner joiner = new StringJoiner(";", "", ";");
 
-  private final Map<String, Object> config;
+  private final JdbcConfig.DuckDbConfig config;
 
   public String buildInitSql() {
     joiner.add("INSTALL iceberg");
@@ -35,7 +35,7 @@ public final class DuckDbExtensions {
     joiner.add("LOAD httpfs");
     ifDiskCache("LOAD cache_httpfs");
 
-    if (asBoolean("use-version-guessing")) {
+    if (config.isUseVersionGuessing()) {
       joiner.add("SET unsafe_enable_version_guessing = true");
     }
 
@@ -43,12 +43,8 @@ public final class DuckDbExtensions {
   }
 
   private void ifDiskCache(String stmt) {
-    if (asBoolean("use-disk-cache")) {
+    if (config.isUseDiskCache()) {
       joiner.add(stmt);
     }
-  }
-
-  private boolean asBoolean(String key) {
-    return (boolean) config.getOrDefault(key, false);
   }
 }
--- a/documentation/docs/configuration-engine/duckdb.md
+++ b/documentation/docs/configuration-engine/duckdb.md
@@ -4,9 +4,11 @@ DuckDB is a vectorized database query engine that excels at analytical queries a
 
 ## Configuration Options
 
-| Key   | Type       | Default          | Description                           |
-|-------|------------|------------------|---------------------------------------|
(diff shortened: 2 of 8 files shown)


sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/server/GraphQLEngineBuilder.java
```diff
@@ -209,18 +209,26 @@ public ResolvedQuery visitSqlQuery(SqlQuery query, Context context) {
...
-              query.updateSQL(
+              query.updateSql(
                   AbstractQueryExecutionContext.addLimitOffsetToQuery(query.getSql(), "?", "?"));
         }
         break;
       default:
         throw new UnsupportedOperationException("Unsupported pagination: " + query.getPagination());
     }
+
+    if (query.getDatabase() == DatabaseType.DUCKDB
```
mbroecheler: Is this the best place to put this? Architecturally, it feels a bit off to me ;-) Maybe we can put this as a "decorator" into the JDBC- or QueryContext?

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/server/GraphQLEngineBuilder.java
```diff
@@ -209,18 +209,26 @@ public ResolvedQuery visitSqlQuery(SqlQuery query, Context context) {
...
+        && query.getTableTraits().contains(TableTrait.GLUE)) {
+
+      var setUnsafeVersionGuess = "SET unsafe_enable_version_guessing = true; ";
```
mbroecheler: I think we have to set this for every catalog, not just Glue. Because whenever you use a catalog, it will write the versions to the catalog instead of the file that DuckDB looks for. The only exception is Hadoop catalog.

So, maybe the Trait should be "FILE_CATALOG" (for "writes all metadata to file") and if that is absent, SET the unsafe thing.

Does this mean that DuckDB checks an enumerates all metadata on every query? There is no caching?

mbroecheler: What does the performance look like?

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/util/DuckDbExtensions.java
```diff
@@ -17,35 +17,50 @@
...
     }
   }
 
-  private void addLoadStatements(StringJoiner joiner) {
-    joiner.add("LOAD iceberg");
-    joiner.add("LOAD httpfs");
-    if (useDiskCache()) {
-      joiner.add("LOAD cache_httpfs");
+  private void ifAws(String stmt) {
+    if (asBoolean("use-aws")) {
```
mbroecheler: Where does this get configured? Is this a manual config or do we infer it from the iceberg configuration?
Same for use-disk-cache.

ferenc-csaky: This is manual, just as `use-disk-cache`:
```json
{
  "engines": {
    "duckdb": {
      "url": "jdbc:duckdb:",
      "config": {
        "use-disk-cache": true,
        "use-aws": true,
        "aws-region": "eu-north-1"
      }
    }
  }
}
```

sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/graphql/server/GraphQLEngineBuilder.java
ferenc-csaky: cache off:
no_cache
cache on:
cache

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/util/DuckDbExtensions.java
mbroecheler: Please make sure those are documented and ping Marvin so this gets added to the default config for cloud compilation. Thanks

PR #1743, opened 2025-12-05
Fix OAuth JWT claims not accessible in `AuthMetadataReader`

velo:
## Summary

- Fixed issue where JWT claims from OAuth/OIDC tokens were not accessible via `user.containsKey()` and `user.get()` in `AuthMetadataReader`
- Changed `OAuth2AuthFactory` to create `JWTAuth` with JWKS instead of `OAuth2Auth`
- This ensures JWT claims are directly accessible on the user principal, matching the behavior expected by `AuthMetadataReader`

## Root Cause

When using `OAuth2AuthHandler`, the JWT access token claims were stored in `user.attributes().getJsonObject("accessToken")` rather than being directly accessible via `user.containsKey(name)`. This caused the server to reject valid tokens with "Required claim missing" errors.

## Changes

- **OAuth2AuthFactory.java**: Now fetches JWKS from OIDC discovery and creates `JWTAuth` instead of `OAuth2Auth`
- **GraphQLServerVerticle.java**: Removed `OAuth2Auth` handling since we now always get `JWTAuth`
- **pom.xml**: Moved `vertx-web-client` from test scope to compile scope (needed for JWKS fetching)

## Test plan

- [x] Unit tests pass (`mvn test -pl sqrl-server/sqrl-server-vertx-base`)
- [ ] Deploy to staging and verify OAuth tokens are accepted


sqrl-server/sqrl-server-vertx-base/pom.xml
```diff
@@ -100,7 +100,6 @@
     <dependency>
       <groupId>io.vertx</groupId>
       <artifactId>vertx-web-client</artifactId>
-      <scope>test</scope>
```
ferenc-csaky, line 103: why change this?

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/auth/OAuth2AuthFactory.java
```diff
@@ -54,27 +61,99 @@ public static Future<AuthenticationProvider> createAuthProvider(
...
+        .send()
+        .map(
+            response -> {
+              if (response.statusCode() != 200) {
+                throw new RuntimeException(
+                    "OIDC discovery failed with status " + response.statusCode());
+              }
+              var jwksUri = response.bodyAsJsonObject().getString("jwks_uri");
+              if (jwksUri == null || jwksUri.isBlank()) {
+                throw new RuntimeException("OIDC discovery document missing jwks_uri");
```
ferenc-csaky: `IllegalStateException` is more appropriate

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/auth/OAuth2AuthFactory.java
```diff
@@ -54,27 +61,99 @@ public static Future<AuthenticationProvider> createAuthProvider(
...
+        .send()
+        .map(
+            response -> {
+              if (response.statusCode() != 200) {
+                throw new RuntimeException(
+                    "JWKS fetch failed with status " + response.statusCode());
+              }
+              var keys = response.bodyAsJsonObject().getJsonArray("keys");
+              if (keys == null || keys.isEmpty()) {
+                throw new RuntimeException("JWKS document has no keys");
```
ferenc-csaky: `IllegalStateException` is more appropriate

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/auth/OAuth2AuthFactory.java
```diff
@@ -54,27 +61,99 @@ public static Future<AuthenticationProvider> createAuthProvider(
...
+      result.add(jwk);
+    }
+    return result;
+  }
+
+  private static void copyIfPresent(JsonObject from, JsonObject to, String key) {
+    if (from.containsKey(key)) {
+      to.put(key, from.getValue(key));
+    }
+  }
```
ferenc-csaky: I'm not a fan of hardcoding those JWKs, but to keep it simple for now, let it be, but at least make it a bit less repetitive:
```java
private static final Set<String> JWK_KEYS = Set.of(
    "kty", "use", "kid", "alg", "n", "e", "x", "y", "crv");

// ...

private static List<JsonObject> convertToJwkOptions(JsonArray keys) {
  var result = new ArrayList<JsonObject>(keys.size());

  for (var obj : keys) {
    var orig = (JsonObject) obj;
    var jwk = new JsonObject();
    for (var entry : orig) {
      if (JWK_KEYS.contains(entry.getKey())) {
        jwk.put(entry.getKey(), entry.getValue());
      }
    }
    result.add(jwk);
  }

  return result;
}
```

sqrl-server/sqrl-server-vertx-base/pom.xml
velo: https://github.com/DataSQRL/sqrl/pull/1743/files#diff-047e0b8133791628667000dc94aa8c2a7756290121b616457386e676112b7667R95

I needed it here

sqrl-testing/sqrl-testing-container/src/test/resources/usecases/jwt-oauth-combined/package.json
ferenc-csaky, line 1: This won't run as an integration test, so in its current form i don't think it should live in that module.

ferenc-csaky:
> FWIW, made image `sqrl-server:0.9.0-alpha-5234924d` and tested, seems to work
> 
> https://datasqrl.slack.com/archives/C06SM7F7RD1/p1764957559476749?thread_ts=1764945754.867669&cid=C06SM7F7RD1

ok, but the test is broken, so this cannot go as is

ferenc-csaky (approved):

PR #1761, opened 2025-12-11
Fix default Iceberg connector config

ferenc-csaky:


sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/IcebergEngine.java
```diff
@@ -74,6 +75,11 @@ protected Map<String, String> getConnectorOptions(String originalTableName, Stri
     var connectorOptions = super.getConnectorOptions(originalTableName, tableId);
 
     if (!isGlueCatalog(connectorOptions)) {
+      if (!connectorOptions.containsKey(ICEBERG_CATALOG_TYPE_KEY)) {
```
mbroecheler: The logic needs to be:
`!connectorOptions.containsKey(ICEBERG_CATALOG_IMPL_KEY) && !connectorOptions.containsKey(ICEBERG_CATALOG_TYPE_KEY)`

Checking for glue only would fail if the user used another catalog impl that isn't glue.

mbroecheler (changes requested):

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/IcebergEngine.java
ferenc-csaky: ~`isGlue` returns false when `ICEBERG_CATALOG_IMPL_KEY ` is not present in the map as well, so adding it here would be redundant.~ Missed the point on first read

mbroecheler (approved):

PR #1795, opened 2026-01-09
Duplication in Subgraph

mbroecheler:
Added a test case that shows duplication in the subgraph. When compiling this test case, you see 6 temporal joins (instead of 2) in the compiled plan. Tried with all settings for `predicate-pushdown-rules`.


ferenc-csaky:
Superseed by https://github.com/DataSQRL/sqrl/pull/1815

PR #1839, opened 2026-02-03
Fix GraphQL schema generation for mutations with nested fields

### LINKED ISSUE #1676: ROW fields in Mutation causes GraphQL cast issue
When a CREATE TABLE defining a mutation has a ROW field (i.e. nested type) this causes issues when generating the GraphQL API if the type is also exposed as a query endpoint because it complains about not being able to cast a ObjectType to an InputType

ferenc-csaky:
## Key Changes
- Removed explicit `GraphQLInputType` cast after `getGraphQLType` call in `GraphqlSchemaUtil`
-  Added new `DAGPlannerTest` case to cover the added logic

Fixes #1676


mbroecheler (approved):
May all of our bugs be this easy to fix ;-)
Nicely done

PR #1852, opened 2026-02-11
Switch JBang from export local to export fatjar with --classpath

velo:
## Summary
- Switch JBang UDF compilation from `export local` to `export fatjar` with `--classpath` pointing to sqrl-cli.jar, so Flink dependencies are provided at compile time without being bundled
- Remove `removeClassPathFromJarManifest()` (no longer needed with fatjar)
- Process only `.java` files with JBang shebang (`///usr/bin/env jbang`), and throw an error if users declare Flink `//DEPS` since Flink is now provided via classpath
- Remove `//DEPS org.apache.flink:flink-table-common` from templates and test resources
- Batch all JBang files into a single `exportFatJar` invocation producing one `jbang-udfs.jar` instead of one JAR per UDF
- Add `complete()` lifecycle hook to `Preprocessor` interface, called after the file tree walk finishes

## Test plan
- [x] All 89 sqrl-cli unit tests pass
- [x] JBangContainerIT passes (116s)
- [x] MetricsContainerIT passes (16.88s)
- [x] Multi-file batching validated via new unit test

### DIFF
--- a/sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
+++ b/sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
@@ -118,7 +118,7 @@ public class JBangPreprocessor extends UdfManifestPreprocessor {
     var classExtendsMatcher = CLASS_EXTENDS_PATTERN.matcher(content);
     var results = classExtendsMatcher.results().toList();
     if (results.isEmpty()) {
-      log.debug(
+      log.info(
           "Skip preprocessing file {}, as it does not contain a 'public class' with an 'extends' statement",
           file);
       return null;


sqrl-cli/src/main/java/com/datasqrl/util/JBangRunner.java
```diff
@@ -39,70 +35,71 @@ public static JBangRunner create() {
...
+
+    cmdLine.addArgument(srcFile.toString());
 
     // TODO: ferenc: add custom PrintStream for stderr of the executor to get JBang error
     var executor = DefaultExecutor.builder().get();
     executor.setExitValue(0);
     executor.execute(cmdLine);
+  }
 
-    removeClassPathFromJarManifest(targetFile);
```
velo, line 66: With export local, JBang produced a JAR with a Class-Path manifest attribute referencing external dependency JARs in a lib/ directory. Those external JARs (Flink etc.) wouldn't be present at runtime in our deployment, so we had to strip the Class-Path entry from the manifest to avoid classloader issues.

With export fatjar, JBang bundles all dependencies inside the JAR itself — no external lib/ directory, no Class-Path manifest entry. The JAR is self-contained, so there's nothing to strip.                                                                                            

And since Flink is now provided via --class-path (not via //DEPS), it's available at compile time but not bundled into the fat JAR. The resulting JAR only contains the user's UDF code and any non-Flink //DEPS they declared.

sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
```diff
@@ -85,7 +120,7 @@ private String parseUdfClassName(Path file, String content) {
     var classExtendsMatcher = CLASS_EXTENDS_PATTERN.matcher(content);
     var results = classExtendsMatcher.results().toList();
     if (results.isEmpty()) {
-      log.warn(
+      log.debug(
```
ferenc-csaky: I'd keep this at least on INFO level, imo this could be useful when some1 just forgot this.

ferenc-csaky (approved):

PR #1893, opened 2026-03-02
Add a last `build` step to CircleCI workflow to match org rules

ferenc-csaky:


.circleci/config.yml
```diff
@@ -604,4 +612,16 @@ workflows:
               ignore: /.*/
           requires:
             - download-maven-dependencies
+      - build:
+          <<: *job-defaults
+          requires:
+            - docker-build-duckdb-extensions
+            - docker-build-base-image
+            - docker-build-mcp-inspector
+            - download-maven-dependencies
```
velo: FWIW, you don't need to depend on those setup images.

PR #1903, opened 2026-03-03
Log full JWT failure stack trace at warn level

velo:
## Summary
- Promotes JWT authentication failure stack trace from `debug` to `warn` level in `JwtFailureHandler`
- Previously, only the root cause message was logged at `warn`, while the full exception chain (which contains the actual rejection reason — signature mismatch, audience, issuer, expiry, etc.) was buried at `debug`
- Now the full stack trace is included in the `warn` log, making JWT failures diagnosable without enabling debug logging

## Test plan
- [x] Unit tests pass (`mvn test -pl sqrl-server/sqrl-server-vertx-base`)
- [ ] Deploy to staging and verify JWT failure logs include full stack traces

### DIFF
--- a/sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/config/ServerConfig.java
+++ b/sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/config/ServerConfig.java
@@ -19,6 +19,7 @@ import static com.datasqrl.graphql.SqrlObjectMapper.MAPPER;
 
 import com.datasqrl.env.EnvVariableNames;
 import com.datasqrl.env.GlobalEnvironmentStore;
+import com.fasterxml.jackson.annotation.JsonGetter;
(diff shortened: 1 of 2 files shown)


sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/auth/JwtFailureHandler.java
```diff
@@ -32,15 +32,21 @@ public void handle(RoutingContext ctx) {
...
-
     if (throwable != null) {
-      log.debug("JWT authentication failure details", throwable);
+      log.warn(
+          "JWT authentication failed for request {} {}: {} (status: {})",
+          ctx.request().method(),
+          ctx.request().path(),
+          errorCause,
+          statusCode,
+          throwable);
```
ferenc-csaky, line 42: This way the throwable is just passed as another argument. If we need to print the stacktrace, you need to use Java string format for the log message and pass only that and the throwable obj to the logger method.

Issue #1921, opened 2026-03-05
Capture INSERT statements in DAG


mbroecheler:
Currently, DataSQRL supports straight up FlinkSQL `INSERT` statements but doesn't capture them as ExportNodes in the DAG, which means these statements (and the entire processing tree leading up to them) is not captured in the graph.

This only affect edge use cases, but it would be nice to capture INSERTs as ExportNodes in the DAG and therefore ensure we capture the lineage. Otherwise, people could consider this a bug because the plan doesn't look right. 

It also means that if the tables we are inserting into are mutation tables, we are not capturing those either and won't plan those. That means the combination of INSERT with mutation tables does not currently work.

PR #1922, opened 2026-03-05
Expanded support for Mutation Tables

mbroecheler:
Improves support for mutation tables, i.e. user defined `CREATE TABLE` statements that do not have a connector which means DataSQRL manages those tables and supports writing to and reading from those tables.

This PR extends support by:
1. Collecting all mutation tables and producing "database" schema for those tables which is printed to the build directory. The goal is to have an authoritative source of all mutations tables in a project/script. This file can then be checked into version control and be used to ensure that future additions/evolutions are backwards compatible. For this, we allow configuring the file under the `script.database` config option in the package.json and - if such a file is present - we do a backwards compatibility check, failing the compilation in case of incompatibility.
2. Generalized mutations and MutationEngine so that we can support Apache Iceberg for mutation table storage. Note, that we do not yet support MutationCoords on the serverside for iceberg which means that vertx can not yet write to those tables.

As we generalize mutation support for engines, we are making a requirement that a mutation table MUST have an `/*+engine(...)*/` hint with the engine that this table is stored in (e.g. kafka, iceberg). This makes it less ambiguous which engine is used. It also makes it more compatible with Flink since newer versions of Flink allow tables without any connector configuration to be used as "schema" tables that are extended with LIKE. This is currently not possible with SQRL. This also allows users to overwrite table config options.
Note, that this breaks backwards compatibility since existing SQRL implementations must add `/*+engine(kafka) */` to mutation tables.

This PR also introduces support for external catalogs and makes the import mechanism more robust.
In addition, it introduces the `maintenance` hint to facilitate table maintenance in streaming projects.

Along the way, I encountered a number of bugs in how we handle mutation tables. Some are fixed in this PR and others have separate tickets: #1921.
Another improvement is in the iceberg connector configuration: moving to v3 and adding upsert.enabled for state tables.

This PR also implements a condensed version of the flink compiled plan which is easier to inspect than the comprehensive version. It is written to `flink-compiled-plan-summary.json` and should be added as an asset to be captured by cloud-backend once this PR lands /cc @velo.

### DIFF
--- a/documentation/docs/sqrl-language.md
+++ b/documentation/docs/sqrl-language.md
@@ -69,7 +69,11 @@ Examples:
 
 SQRL understands the complete Flink SQL `CREATE TABLE` syntax, but distinguishes between **internal** and **external** source tables.
 External source tables are standard Flink SQL tables that connect to an external data source (e.g. database or Kafka cluster).
-Internal tables connect to a data source that is managed by SQRL (depending on the configured `log` engine, e.g. a Kafka topic) and exposed for data ingestion in the interface.
+Internal tables connect to a data source that is managed by SQRL and exposed for data ingestion in the interface.
+
+:::important
+Internal tables support `kafka` and `iceberg` engines.
+:::
 
 | Feature                       | Internal source (managed by SQRL)                               | External Source (connector) |
 |-------------------------------|-----------------------------------------------------------------|-----------------------------|
@@ -79,7 +83,7 @@ Internal tables connect to a data source that is managed by SQRL (depending on t
 | Watermark spec                | **generated**                                                   | **required**                |
 | Primary key                   | *Unenforced* upsert semantics                                   | Same as Flink               |
 
-Example (internal):
+### Internal Example
 
 ```sql
 /*+ engine(kafka) */
@@ -92,7 +96,7 @@ CREATE TABLE Customer (
 );
 ```
 
-Example (external):
+### External example
 
 ```sql
 CREATE TABLE kafka_json_table (


mbroecheler:
@ferenc-csaky Apologies, this got a little beefy, but since I was already refactoring how we handle CREATE TABLE statements, I decided to also fix #1924.

This requires a 0.10.0 release.

sqrl-cli/src/main/java/com/datasqrl/compile/DagWriter.java
```diff
@@ -40,6 +42,7 @@ public class DagWriter {
   public static final String EXPLAIN_VISUAL_FILENAME = "pipeline_visual.html";
   public static final String EXPLAIN_JSON_FILENAME = "pipeline_explain.json";
   public static final String FULL_SOURCE_FILENAME = "pipeline_source.sqrl";
+  public static final String DATABASE_FILENAME = "pipeline_database.json";
```
ferenc-csaky: This is mutation-specific, and that's not obvious from its name, and even its content is `tables`, so maybe give the file a more specific name, e.g. `pipeline_mutation_database.json`?

sqrl-cli/src/main/java/com/datasqrl/packager/FilePreprocessingPipeline.java
```diff
@@ -75,6 +75,7 @@ private static Set<String> getCopyExtensions() {
     copyExtensions.add(SqrlConstants.SQRL_EXTENSION);
     copyExtensions.add(SqrlConstants.SQL_EXTENSION);
     copyExtensions.add(SqrlConstants.GRAPHQL_EXTENSION);
+    copyExtensions.add(SqrlConstants.CONFIG_EXTENSION);
```
ferenc-csaky, line 78: This is simply `json`. What config this refers to? I guess not `package.json` cause then why don't we needed it before. And in that case. I think this will produce undesired behavior, processing all json files.

sqrl-cli/src/main/java/com/datasqrl/compile/DagWriter.java
mbroecheler: `pipeline_mutation_database.json` is great.

sqrl-cli/src/main/java/com/datasqrl/packager/FilePreprocessingPipeline.java
mbroecheler: For the package.json we do special preprocessing in the Packager to generate the consolidate json. I needed to have other config files copied over (like the mutation database) so they are available at compile-time. This is the best I could come up with.
Yes, it does copy all config files but that seems harmless. And it sets us up to have potentially other config files that we copy over automatically in the future.

PR #1923, opened 2026-03-05
Rename CircleCI workflow to build

velo:
## Summary
- Rename CircleCI workflow from `ci` to `build` so it reports as a single `build` status check on GitHub — matching the org-wide branch protection requirement
- Remove the `.github/workflows/build.yml` polling hack that was working around this by watching individual CircleCI job statuses

## Test plan
- [ ] Verify CircleCI reports `ci/circleci: build` as workflow-level status on the PR
- [ ] Confirm branch protection sees it as the required `build` check


.circleci/config.yml
```diff
@@ -549,7 +549,7 @@ jobs:
               .
 
 workflows:
-  ci:
+  build:
```
velo, line 552: I don't know if this was attempted before, think worth a shot.

PR #1934, opened 2026-03-09
Remove deployment validation from packageSchema.json

velo:
## Summary
- Remove `deployment` property definitions from flink, vertx, and postgres engine schemas in `packageSchema.json`
- Change `additionalProperties` from `false` to `true` for these engines, allowing deployment config to pass through without validation (matching how kafka engine already works)
- Delete 4 test resource files and remove corresponding test cases that validated deployment-specific constraints

## Rationale
The compiler never reads or uses deployment config values — they are consumed by external tooling (e.g., sqrl-k8s). Moving validation out of the compiler removes unnecessary coupling and lets deployment tooling evolve independently.

## Test plan
- [x] `PackageJsonSchemaTest` passes (20 tests, 0 failures)
- [x] All `sqrl-planner` tests pass (169 tests, 0 failures)
- [x] Valid deployment test files (`validFlinkDeployment.json`, `validPostgresDeployment.json`, `validVertxDeployment.json`, `validFullDeployment.json`) still parse without error


sqrl-planner/src/main/resources/jsonSchema/packageSchema.json
```diff
@@ -186,56 +186,9 @@
             "config": {
               "type": "object",
               "minProperties": 1
-            },
-            "deployment": {
```
ferenc-csaky, line 190: Maybe we wanna keep `deployment`, and only be lenient about things inside that key?

velo: Ow, nice catch, I changed it a wrong level 🤦‍♂️

ferenc-csaky (approved):

PR #1940, opened 2026-03-10
Fixes bug in sub-query evaluation

mbroecheler:
Previously, sub-query rewriting only worked for trivial base tables. Added test case. Changed test infrastructure a bit so it is easier to run external datasqrl projects for debugging.


sqrl-testing/sqrl-testing-integration/src/test/java/com/datasqrl/ExternalUseCaseCompileTest.java
```diff
@@ -19,11 +19,11 @@
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
-@Disabled
+// @Disabled
```
ferenc-csaky: This needs to stay disabled.

ferenc-csaky (changes requested):

ferenc-csaky (approved):

PR #1943, opened 2026-03-12
Restrict `CREATE CATALOG` statements to separate scripts not imported inline

ferenc-csaky:
## Key Changes
- Intercept `CREATE CATALOG` statements and validate that they are done in a separate script
- Add test case to the introduced validation logic
- Document `CREATE CATALOG` limitations

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/planner/SqlScriptPlanner.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/planner/SqlScriptPlanner.java
@@ -553,7 +553,7 @@ public class SqlScriptPlanner {
       } else if (node instanceof SqlDropTable || node instanceof SqlDropView) {
         errors.fatal(
             "Removing tables is not supported. The DAG planner automatically removes unused tables.");
-      } else if (node instanceof SqlCreateCatalog && scriptContext.generateAccess()) {
+      } else if (node instanceof SqlCreateCatalog && scriptContext.isRootContext()) {
         errors.fatal(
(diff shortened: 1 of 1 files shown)


sqrl-planner/src/main/java/com/datasqrl/planner/SqlScriptPlanner.java
```diff
@@ -1167,12 +1171,12 @@ public HintsAndDocs dropHints() {
   }
 
   private record ScriptContext(
-      ModuleLoader moduleLoader, String databaseName, boolean generateAccess) {
+      ModuleLoader moduleLoader, String databaseName, boolean generateAccess, boolean root) {
```
mbroecheler: I'm wondering, isn't root and generate access the same thing? Meaning shouldn't it be:

public boolean generateAccess() { return root; }

instead of a separate field. We generate access only for tables that are defined in the main script (or inline imports).

ferenc-csaky: TBH I did not tracked down what exactly `generateAccess` is used for. :)

ferenc-csaky: ~I think that name could be more explicit, e.g. `rootContext` instead, and then I won't bump into this 6 months from now again. WDYT?~

I see that by how everything else uses it, it makes sense to call it `generateAccess`, I'll just document this for my dummy self. :)

PR #1963, opened 2026-03-24
feat: Generate split SQL deployment artifacts for multi-batch projects

ferenc-csaky:
## Key Changes
- Generate individual split SQL deployment artifacts in case of a multi-batch project
- These extra artifacts will only be generated when there are more than 1 batch
- Every split will only contain the previous and current DDLs, so for example we don't register unnecessary tables that are defined in an upcoming batch

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/planner/FlinkPhysicalPlan.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/planner/FlinkPhysicalPlan.java
@@ -23,7 +23,9 @@ import com.datasqrl.engine.stream.flink.sql.RelToFlinkSql;
 import com.datasqrl.planner.tables.FlinkConnectorConfigWrapper;
 import com.datasqrl.planner.util.CompiledPlanCondenser;
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ListMultimap;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


sqrl-planner/src/main/java/com/datasqrl/planner/FlinkPhysicalPlan.java
```diff
@@ -94,6 +124,8 @@ public static class Builder {
     private final Set<String> formats = new HashSet<>();
     private final Set<String> fullyResolvedFunctions = new HashSet<>();
     private final List<List<RichSqlInsert>> statementSets = new ArrayList<>();
+    private final List<List<String>> flinkSqlBatched = new ArrayList<>();
+    private final List<List<String>> flinkSqlNoFunctionsBatched = new ArrayList<>();
```
velo: Shouldn't we use something like a `ArrayListMultimap`?!

Then you can just `flinkSqlBatched.put(idx, sql)`

sqrl-planner/src/main/java/com/datasqrl/planner/FlinkPhysicalPlan.java
```diff
@@ -127,12 +172,13 @@ public void add(List<? extends SqlNode> nodes, List<String> nodeSqls) {
...
       for (int i = 0; i < nodes.size(); i++) {
-        add(nodes.get(i), nodeSqls.get(i));
+        add(nodes.get(i), nodeSqls.get(i), i);
       }
     }
 
-    public void add(SqlNode node, String nodeSql) {
+    private void add(SqlNode node, String nodeSql, int batchIdx) {
       flinkSql.add(nodeSql);
+      flinkSqlBatched.get(batchIdx).add(nodeSql);
```
velo: My brain got hurt trying to understand why this works =)

sqrl-testing/sqrl-testing-integration/src/test/resources/snapshots/com/datasqrl/UseCaseCompileTest/multi-batch-compile-package.txt
```diff
@@ -84,4 +90,10 @@ INSERT INTO `default_catalog`.`default_database`.`SinkB_ex2`
...
   FROM `default_catalog`.`default_database`.`Src`
  ;
- END
+ END;
+EXECUTE STATEMENT SET BEGIN
+INSERT INTO `TableC`
+SELECT *
+ FROM `Src`
+;
+END
```
velo, line 233: We probably want to update the snapshot collector so it collects the 2 segments.

PR #1967, opened 2026-03-26
fix: Fail early with proper error message when mutation engine hint is missing

ferenc-csaky:
## Key Changes
- Fail early with descriptive error message when internal `CREATE TABLE` is missing a mutation engine hint
- Add test case to cover the added logic


sqrl-testing/sqrl-testing-integration/src/test/resources/snapshots/com/datasqrl/DAGPlannerTest/mutationHintMissing-fail.txt
```diff
@@ -0,0 +1,5 @@
+[FATAL] Mutation engine hint is required for internal CREATE TABLE statements
```
velo: Any change we can get an explanation of what is missing?!

I read the error, but I'm not sure what it mean =(

velo (approved):

PR #1988, opened 2026-04-01
chore: Refactor AbstractFullUseCaseTest to JUnit extension

### LINKED ISSUE #1987: Refactor AbstractFullUseCaseTest
The class owns static `TestContainerHook` management (Kafka/Postgres via the visitor
pattern), Flink mini-cluster coordination, per-test `TestExtension` instantiation, and environment
variable injection — all via inheritance.

mateczagany:
- Move TestContainerHook-related logic from `AbstractFullUseCaseTest` to a new JUnit extension `FullPipelineContainerExtension` 
 - Add `FullUseCaseRunner` utility class with the same helper methods that `AbstractFullUseCaseTest` to execute the tests

Closes #1987

### DIFF
--- a/sqrl-testing/sqrl-testing-integration/src/test/java/com/datasqrl/ExternalUseCaseIT.java
+++ b/sqrl-testing/sqrl-testing-integration/src/test/java/com/datasqrl/ExternalUseCaseIT.java
@@ -16,6 +16,7 @@
 package com.datasqrl;
 
 import com.datasqrl.engines.FullPipelineContainerExtension;
+import com.datasqrl.engines.TestContainersForTestGoal.TestContainerHook;
 import com.datasqrl.tests.DuckdbTestExtension;
 import com.datasqrl.tests.IcebergTestExtension;
 import com.datasqrl.tests.SnowflakeTestExtension;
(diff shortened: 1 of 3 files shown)


sqrl-testing/sqrl-testing-integration/src/test/java/com/datasqrl/ExternalUseCaseIT.java
```diff
@@ -31,13 +32,18 @@
...
+  IcebergTestExtension.class,
+  SnowflakeTestExtension.class
+})
+public class ExternalUseCaseIT {
 
   @ParameterizedTest
   @MethodSource("externalUseCaseProvider")
   void testCase(UseCaseParam param) {
-    fullUseCaseTest(param);
+    FullUseCaseRunner.run(param, FullPipelineContainerExtension.getHook());
```
ferenc-csaky: AFAIK if the extension implements `ParameterResolver` and we set the hook there properly, we can just have it as an added param on any `@ParameterizedTest`, and we can eliminate this static call, that's only working cause the extension callback sets it right, but I'd consider that fragile.

mateczagany: You are right, it makes a lot more sense like this, especially with sharding

PR #2007, opened 2026-04-13
fix: Write mutation records with keys properly to Kafka

ferenc-csaky:
## Key Changes
- Wire in the `Flink` primary key col names to `KafkaMutationCoords`
- In `MutationConfigurationImpl`, when we put together the Kafka record to be sent, we take the key columns into account, and also send it with the value if it exists

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/graphql/GraphqlModelGenerator.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/graphql/GraphqlModelGenerator.java
@@ -106,7 +106,6 @@ public class GraphqlModelGenerator extends GraphqlSchemaWalker {
       FieldDefinition atField, TypeDefinitionRegistry registry, MutationTable mutation) {
 
     var computedCols = mutation.getComputedColumns();
-    var keysCols = mutation.getTableBuilder().getPrimaryKey().orElse(List.of());
     var returnList = GraphqlSchemaUtil.isListType(atField.getType());
 
     if (mutation.getCreateTable() instanceof KafkaLogEngine.Table mutationTopic) {
@@ -115,7 +114,7 @@ public class GraphqlModelGenerator extends GraphqlSchemaWalker {
               atField.getName(),
               returnList,
               mutationTopic.topicName(),
-              Set.copyOf(keysCols),
+              mutationTopic.messageKeys(),
               computedCols,
               mutation.getInsertType() == MutationInsertType.TRANSACTION,
               Map.of()));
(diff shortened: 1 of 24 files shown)


sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/MutationConfigurationImpl.java
```diff
@@ -129,6 +139,20 @@ private List<Map> getEntries(
       entries = List.of((Map) argument);
     }
 
+    // Construct keys if necessary (null if no key cols)
+    var keys = new Map[entries.size()];
+    if (!keyColumns.isEmpty()) {
+      for (int i = 0; i < entries.size(); i++) {
+        var filtered = new LinkedHashMap();
+        for (var key : keyColumns) {
+          if (entries.get(i).containsKey(key)) {
```
mateczagany, line 147: This probably won't maintain the order of a composite primary key, but I'm not sure if that's a concern.

sqrl-planner/src/main/java/com/datasqrl/graphql/GraphqlModelGenerator.java
```diff
@@ -104,15 +104,19 @@ protected void visitSubscription(
   @Override
   protected void visitMutation(
       FieldDefinition atField, TypeDefinitionRegistry registry, MutationTable mutation) {
-    var computedColumns = mutation.getComputedColumns();
+
+    var computedCols = mutation.getComputedColumns();
+    var keysCols = mutation.getTableBuilder().getPrimaryKey().orElse(List.of());
```
mbroecheler: This is not necessary because we already set this on the KafkaLogEngine.Table below.
Instead, use `mutationTopic.messageKeys()`. That also covers partition keys.

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/MutationConfigurationImpl.java
mbroecheler: it only needs to preserve the order of keyColumns.

mbroecheler (changes requested):
One minor change on where you get the keys from.
Otherwise looks good. Thanks!

It might be worth considering adding a test for PARTITIONED BY to see if that works. The code is all there in this PR.

mateczagany (approved):

PR #2016, opened 2026-04-14
feat: Support NULL as GraphQL API argument

### LINKED ISSUE #2005: Support NULL as argument
Null arguments in graphql API results in “could not determine type” error in database.

This happens when you define a function with nullabe argument such as:
```
MyFunction(arg BIGINT) := SELECT * FROM MyTable WHERE (:arg IS NULL OR column = :arg);
```
Instead, we need to set such arguments using:
`preparedStatement.setNull(1, Types.VARCHAR);`

ferenc-csaky:
Resolves #2005


mbroecheler (approved):
That looks clean to me. Thanks.
Did not have the time to review all snapshot changes.

ferenc-csaky:
@mbroecheler Snapshot changes only contain the newly introduced `sqlType` field for any case that contains a `ArgumentParameter`.

PR #2019, opened 2026-04-16
feat: Introduce Kafka health check in Vert.x

velo:
## Summary

When the Kafka broker becomes unreachable (e.g. DNS resolution failure), the `/health` endpoint kept returning 204 and Kubernetes never restarted the pod. Mutation HTTP requests continue to fail naturally via the existing `cf::completeExceptionally` propagation when the producer eventually times out — that part already worked.

This PR adds a self-probing Kafka health check:

- New `KafkaHealthTracker` owns a dedicated `KafkaAdminClient` configured with a short request/API timeout (5s default). A periodic timer (10s default) calls `listTopics()` and caches the result. The `/health` endpoint reads the cached state, so it never blocks.
- `HttpServerVerticle` registers a `kafka` health-check procedure when `kafkaMutationConfig` is set; the tracker is added to `closeables` for orderly shutdown.
- Self-healing: once the broker is reachable again, the next probe flips the cached state back to healthy without depending on user mutation traffic.
- No changes to producer timeouts — mutation send semantics are unchanged.

## Test plan

- [x] `mvn test` (full suite, 0 failures)
- [x] New `KafkaHealthTrackerTest` — points the tracker at an unreachable broker and verifies it transitions to unhealthy with a non-blank error within ~10s
- [ ] Manual smoke: deploy to a staging cluster, kill broker DNS, verify `/health` flips to 503 within ~15s


sqrl-server/sqrl-server-vertx-base/src/test/java/com/datasqrl/graphql/kafka/KafkaHealthTrackerTest.java
```diff
@@ -0,0 +1,103 @@
...
+  void tearDown() {
+    vertx.close();
+  }
+
+  @Test
+  void givenUnreachableBroker_whenProbeFires_thenTrackerReportsUnhealthy() throws Exception {
+    var config = Map.of(BOOTSTRAP_SERVERS_CONFIG, "127.0.0.1:1");
+    try (var tracker = new KafkaHealthTracker(vertx, config, 200L, 500L)) {
+      var deadline = System.currentTimeMillis() + 10_000L;
+      while (tracker.isHealthy() && System.currentTimeMillis() < deadline) {
```
mateczagany: Awaitility would be nicer here I think: 

```
      Awaitility.waitAtMost(Duration.ofSeconds(10))
              .until(() -> !tracker.isHealthy());
```

sqrl-server/sqrl-server-vertx-base/src/test/java/com/datasqrl/graphql/kafka/KafkaHealthTrackerTest.java
```diff
@@ -0,0 +1,103 @@
...
+      assertThat(tracker.lastError()).isNotBlank();
+    }
+  }
+
+  @Test
+  void givenUnhealthyTracker_whenHealthEndpointHit_thenReturns503() throws Exception {
+    var config = Map.of(BOOTSTRAP_SERVERS_CONFIG, "127.0.0.1:1");
+    try (var tracker = new KafkaHealthTracker(vertx, config, 200L, 500L)) {
+      var deadline = System.currentTimeMillis() + 10_000L;
+      while (tracker.isHealthy() && System.currentTimeMillis() < deadline) {
```
mateczagany: Here as well:
```
      Awaitility.waitAtMost(Duration.ofSeconds(10))
              .until(() -> !tracker.isHealthy());
```

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/graphql/kafka/KafkaHealthTracker.java
```diff
@@ -0,0 +1,89 @@
...
+    this(vertx, baseKafkaConfig, DEFAULT_PROBE_INTERVAL_MS, DEFAULT_PROBE_TIMEOUT_MS);
+  }
+
+  public KafkaHealthTracker(
+      Vertx vertx, Map<String, String> baseKafkaConfig, long probeIntervalMs, long probeTimeoutMs) {
+    this.vertx = vertx;
+    var probeConfig = new HashMap<>(baseKafkaConfig);
+    probeConfig.put(REQUEST_TIMEOUT_MS_CONFIG, String.valueOf(probeTimeoutMs));
+    probeConfig.put(DEFAULT_API_TIMEOUT_MS_CONFIG, String.valueOf(probeTimeoutMs));
+    this.adminClient = KafkaAdminClient.create(vertx, probeConfig);
```
mateczagany: This is the only place where `KafkaAdminClient` is used in the repository, and `listTopics()` might require permissions that's not 100% obvious to the user that they should grant.

Maybe we could use `KafkaProducer#partitionsFor` for this check, so no additional permissions need to be given to the application, but for that you would need to know the topics in advance, I'm not sure if that's possible

ferenc-csaky: AFAIK the only risk regarding `KafkaAdminClient` is `listTopics()` requires `Describe` on the `Cluster` resource. If the cluster uses authorization and the principal was provisioned narrowly (e.g. only `Write` / `Describe` on specific topics), the probe will fail permanently with an auth ex.

This should be mitigated with `describeTopics`, that will have one of the mutation topics, that should always have the proper permissions.

mateczagany (approved):

ferenc-csaky:
I will merge this, when the container tests properly support having a Kafka (Redpanda) cluster in a serparate change.

ferenc-csaky:
This will need some extra love, cause:

1. `mutationConfig` is always present in the `server-config.json`, so it's not a good indicator to decide if we need Kafka health tracker or not
2. It's also possible that a project has subscriptions, but not mutations, so we should apply the tracker in both cases

My idea is to iterate over the `RootGraphqlModel`s and pick whichever we find first, and attach the relevant config map to the tracker.

ferenc-csaky (approved):

PR #2022, opened 2026-04-20
fix: Defined operations are ignored if GraphQL schema is inferred

### LINKED ISSUE #2021: Defined operations are ignored if GraphQL schema is inferred
When no explicit GraphQL schema is provided and the schema is inferred, any user-defined operation files are silently ignored and dropped from the compiled output.

mateczagany:
Fixes #2021 

- Preserve user-defined operations when GraphQL schema is inferred by adding a new, `defaultOperations` field
- Add unit tests

### DIFF
--- a/sqrl-planner/src/test/java/com/datasqrl/config/GraphqlSourceLoaderTest.java
+++ b/sqrl-planner/src/test/java/com/datasqrl/config/GraphqlSourceLoaderTest.java
@@ -17,230 +17,198 @@ package com.datasqrl.config;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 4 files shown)


sqrl-planner/src/test/java/com/datasqrl/config/GraphqlSourceLoaderTest.java
```diff
@@ -0,0 +1,246 @@
...
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessageContaining("does-not-exist.graphql");
+  }
+
+  private static ScriptFiles scriptFiles(
+      @Nullable String graphql, List<String> operations, List<ScriptApiConfig> apiConfigs) {
+    return new ScriptFiles(stubPackageJson(graphql, operations, apiConfigs));
+  }
+
+  private static PackageJson stubPackageJson(
```
ferenc-csaky: Let's use mockito instead of defining these.

ferenc-csaky (commented):
We had an offline discussion to shape `GraphqlSourceLoader` a bit to make its flow cleaner. Will check after that update.

ferenc-csaky (approved):
LGTM

PR #2055, opened 2026-05-07
feat: Table type specific PG on conflict behavior

ferenc-csaky:
## Key Changes
- Only relevant logic change in `PostgresJdbcEngine`
  - Ignore conflicting records for `STREAM` tables
  - Use the `ROWTIME` field as `on-conflict.timestamp-column` for `STATE` tables if there is a `ROWTIME` field in the schema
  - Otherwise, default on-conflict behavior
- Refactored `ConnectorConf.Context`, eliminmated unused variables
- Refactored `AbstractJDBCEngine#getConnectorOptions` siganture, now it takes the `ContextBuilder` object so it will be more flexible if we introduce variables
- Adapted snapshots

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresJdbcEngine.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresJdbcEngine.java
@@ -15,13 +15,15 @@
  */
 package com.datasqrl.engine.database.relational;
 
+import static com.datasqrl.flinkrunner.connector.postgresql.jdbc.SqrlPostgresOptions.SINK_ON_CONFLICT;
+import static com.datasqrl.flinkrunner.connector.postgresql.jdbc.SqrlPostgresOptions.SINK_ON_CONFLICT_COLUMN;
+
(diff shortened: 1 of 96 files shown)


mateczagany (approved):

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresJdbcEngine.java
```diff
@@ -53,4 +60,36 @@ protected DatabaseType getDatabaseType() {
...
+    var tableType = tableAnalysis.getType();
+    if (tableType.isStream()) {
+      mutableOptions.put(
+          SqrlPostgresOptions.SINK_ON_CONFLICT.key(), OnConflictAction.IGNORE.name());
+
+      return mutableOptions;
+    }
+
+    if (tableType.isState()) {
+      for (var field : tableAnalysis.getRowType().getFieldList()) {
```
mbroecheler: Please use TableAnalysis#getRowTime() for this. it returns the index of the rowtime field.

mbroecheler (changes requested):
One minor change for DRYness.

PR #2073, opened 2026-05-13
fix: Rework comment parsing in `SqlScriptStatementSplitter`

ferenc-csaky:
## Key Changes
- Reworked `SqlScriptStatementSplitter` to parse line comments, block comments, doc comments, SQL hints, and single-quoted string literals in one stateful pass.
- Removed regex-based block comment stripping so `/* ... */` inside string literals is preserved correctly.
- Preserved Java-style doc comments `/** ... */` and SQL hints `/*+ ... */`, while removing regular block comments and `--` line comments outside string literals.
- Added regression tests for multiline string literals, doc comments with quotes/comment markers, block comments inside/outside literals, multiline block comments, and exact SQL hint detection.

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/planner/parser/SqlScriptStatementSplitter.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/planner/parser/SqlScriptStatementSplitter.java
@@ -20,7 +20,6 @@ import com.google.common.base.CharMatcher;
 import com.google.common.base.Preconditions;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.regex.Pattern;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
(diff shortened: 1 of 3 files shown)


mateczagany (approved):

mateczagany:
Looks good! But while reviewing this, I've found another edge case that would break the splitter, by `removeBlockComments` removing the content of the string literal:
```
SELECT '/* not a comment */' AS val;
```

ferenc-csaky:
> Looks good! But while reviewing this, I've found another edge case that would break the splitter, by `removeBlockComments` removing the content of the string literal:
> 
> ```
> SELECT '/* not a comment */' AS val;
> ```

Yep, you have a good eye :) I'm already on that, adding some more test cases, and probably doing a little bigger refactor in the splitter.

PR #2075, opened 2026-05-14
ci: Add `datasqrl-examples` bump and validation workflows

velo:
## Summary
- Add a manual GitHub Actions workflow to validate `datasqrl-examples` against a selected SQRL CLI image.
- Default the manual validation image to `datasqrl/cmd:dev`, while still allowing maintainers to pass snapshot images like `ghcr.io/datasqrl/cmd:pr-123`.
- Dispatch `DataSQRL/datasqrl-examples` `build.yml`, locate the downstream run by `correlation_id`, and wait for it to finish.
- Keep `datasqrl-examples` validation detached from CircleCI instead of running it as a CircleCI job.
- Allow CircleCI to publish PR images to GHCR with `pr-*` tags so they can be manually validated by the examples workflow.
- Add a release/manual workflow that opens a PR in `datasqrl-examples` to bump its `SQRL_VERSION`.
- Use the dedicated `EXAMPLES_BOT_PAT` secret for cross-repo examples operations.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -400,10 +400,12 @@ jobs:
           name: Skip image generation
           command: |
             set -x
-            
-            # Push to Docker Hub for tagged releases and main=:dev.
-            if [ "$IS_RELEASE" != "true" ] && [ "$TAG_SUFFIX" != "dev" ]; then
-              echo "Skipping image publishing for TAG_SUFFIX=$TAG_SUFFIX (only tagged releases and main=:dev are published)."
+
+            # Push to Docker Hub for tagged releases (datasqrl/*) and main=:dev (datasqrl/*).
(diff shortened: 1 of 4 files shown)


WellMafra (approved):

ferenc-csaky (changes requested):
I don't think this should be executed for every PR or commit that makes it into `main`.

Ideally that would be nice, but IMO in real life that will be just another annoying step we just skip most of the time, and will be red until we release anyways.

WDYT?

velo:
> I don't think this should be executed for every PR or commit that makes it into `main`.
> 
> Ideally that would be nice, but IMO in real life that will be just another annoying step we just skip most of the time, and will be red until we release anyways.
> 
> WDYT?

well, if we don't run this on every PR, we will only know it's broken after someone manually checks

ferenc-csaky (approved):

PR #2086, opened 2026-05-21
feat: Make Kafka topic numPartitions and replicationFactor configurable

velo:
## Summary
- Adds `engines.kafka.num-partitions` (default `1`) and `engines.kafka.replication-factor` (default `3`) to `default-package.json`
- `KafkaLogEngine` reads them from the engine config and propagates them to both subscription/mutation topics
-  The test-runner topics are restricted to 1 partition and 1 replication factor


sqrl-planner/src/main/java/com/datasqrl/engine/log/kafka/KafkaLogEngine.java
```diff
@@ -340,7 +354,18 @@ public EnginePhysicalPlan plan(MaterializationStagePlan stagePlan) {
 
     var testRunnerTopics =
         testRunnerConfig.getCreateTopics().stream()
-            .map(topicName -> new NewTopic(topicName, topicName))
```
ferenc-csaky, line 343: I'd keep the short ctor in the record.

sqrl-planner/src/main/java/com/datasqrl/engine/log/kafka/KafkaLogEngine.java
```diff
@@ -75,6 +75,10 @@ public class KafkaLogEngine extends ExecutionEngine.Base implements LogEngine {
   public static final String UPSERT_FORMAT = "upsert-%s";
 
   public static final String DEFAULT_TTL_KEY = "retention";
+  public static final String NUM_PARTITIONS_KEY = "num-partitions";
+  public static final String REPLICATION_FACTOR_KEY = "replication-factor";
+  public static final int DEFAULT_NUM_PARTITIONS = 1;
+  public static final short DEFAULT_REPLICATION_FACTOR = 3;
```
ferenc-csaky: I think we can omit these. The `getSetting` will have those values no matter what, cause they're part of the `default-package.json`. The rest of the configs are also work the same way.

ferenc-csaky (changes requested):
Logic itself LG, added 2 improvement suggestions.

ferenc-csaky (approved):

PR #2098, opened 2026-05-26
feat: DuckDB S3 credential-chain support for IRSA-backed Iceberg reads

velo:
Reopens/supersedes #2096 (that PR's branch was deleted so it can't be reopened in place).

## Problem

When the Vert.x server reads an S3-backed Iceberg table through the **DuckDB** engine from an in-cluster pod (EKS **IRSA**), the query fails with `HTTP 403 Forbidden / AccessDenied — No credentials are provided`.

DuckDB's `httpfs` only auto-detects **env-var** credentials (`AWS_ACCESS_KEY_ID`/`SECRET`/`SESSION_TOKEN`) or an explicit credential secret. It does **not** use the projected web-identity token that backs IRSA, so on an IRSA-only pod it sends an **anonymous** request → 403. (Flink writes succeed because Flink's S3 layer does the web-identity exchange.)

This was proven directly inside a running server pod: anonymous `iceberg_scan('s3://…')` → `403 No credentials`; after adding the credential-chain secret below, the same read returned **1,810,576 rows**.

## Change

Opt-in `use-credential-chain` flag on the DuckDB engine. When enabled, the connection init SQL adds:

```sql
LOAD aws;
CREATE OR REPLACE SECRET sqrl_s3_credential_chain (TYPE S3, PROVIDER credential_chain);
```

`PROVIDER credential_chain` (with **no explicit `CHAIN`**) delegates to the AWS SDK default provider chain, which includes the web-identity provider used by IRSA — so DuckDB obtains temporary creds from the projected service-account token, exactly like Flink.

> Note: an explicit `CHAIN 'env;config;sts;…'` does **not** work — DuckDB maps `sts` to AssumeRole and rejects it without an `ASSUME_ROLE_ARN`. The default chain is required. (This is the correction vs the original #2096.)

The flag defaults to `false`, preserving today's behavior.

## Files

- `DuckDbExtensions.java` — emit `LOAD aws` + default-chain `CREATE SECRET` when the flag is set; `@VisibleForTesting buildInitSql(extensionDir)` overload for unit testing.
- `JdbcConfig.DuckDbConfig` — new `use-credential-chain` boolean.
- `packageSchema.json` — register the new key.
- `Dockerfile.duckdb-extensions` — `INSTALL aws` so the extension is bundled (the `credential_chain` provider requires it; confirmed it is not bundled today).
- `documentation/docs/configuration-engine/duckdb.md` — document the flag + IRSA rationale.
- `DuckDbExtensionsTest.java` — covers default vs credential-chain init SQL, ordering, all-flags, and that no explicit `CHAIN` is emitted.

## Validation

- `DuckDbExtensionsTest` — 4 tests green (JDK 17).
- The `credential_chain` secret was validated live in a server pod (IRSA-only): private-bucket `iceberg_scan` returned 1.8M rows with the secret, 403 without it.
- Downstream: `IcebergDeploymentIT` in cloud-compilation has the full datagen→Iceberg(Glue/S3FileIO/`s3://`)→Flink-write path green; this PR closes the read-side gap. Setting `use-credential-chain: true` on the duckdb engine turns that IT's GraphQL read green.

Draft until a maintainer confirms the secret name + bundling the `aws` extension.

### DIFF
--- a/documentation/docs/configuration-engine/duckdb.md
+++ b/documentation/docs/configuration-engine/duckdb.md
@@ -10,6 +10,7 @@ DuckDB is a vectorized database query engine that excels at analytical queries a
 | `memory-limit`         | **string**  | -                | Sets DuckDB's `memory_limit`, for example `"8GB"`                              |
 | `use-disk-cache`       | **boolean** | `false`          | Install and load `cache_httpfs` extension                                      |
 | `use-version-guessing` | **boolean** | `false`          | Sets `unsafe_enable_version_guessing` flag to be able to read uncommitted data |
+| `use-credential-chain` | **boolean** | `false`          | Load the `aws` extension and create an S3 secret backed by the AWS credential provider chain |
 
 ## Example Configuration
 
@@ -31,6 +32,7 @@ DuckDB is a vectorized database query engine that excels at analytical queries a
 - Ideal for local development and testing of analytical workloads
 - Excellent performance on analytical queries with vectorized execution
 - Can read Iceberg tables directly without additional infrastructure
+- Enable `use-credential-chain` when reading S3-backed Iceberg from an environment that supplies credentials through the AWS provider chain rather than static keys — e.g. EKS IRSA, where the pod only has a projected web-identity token. Plain `httpfs` does not perform the web-identity exchange, so without this flag S3 reads fail with `HTTP 403`.
 - Supports both in-memory and persistent database modes
 - Perfect for prototyping before deploying to cloud query engines like Snowflake
 - Lightweight alternative to larger analytical databases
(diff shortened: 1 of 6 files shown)


mbroecheler:
@velo This change makes sense to me.
@ferenc-csaky One thing we want to look into is how we support the different clouds with this one to load things dynamically. 

Note, that this is currently blocked on a DuckDB fix.

mateczagany (approved):

documentation/docs/configuration-engine/duckdb.md
```diff
@@ -31,6 +32,21 @@ DuckDB is a vectorized database query engine that excels at analytical queries a
...
+connection is opened and are **never refreshed for the life of that connection**. Provider-chain
+credentials are usually temporary — an IRSA web-identity session typically lasts an hour — so a
+long-running server can start failing S3 reads with `HTTP 403` once they expire, and keep failing
+until the pool opens a new physical connection and re-runs the init SQL.
+
+This cannot be fixed from SQRL alone: DuckDB's `httpfs` does not refresh credentials at every call
+site that needs them ([duckdb/duckdb-httpfs#165](https://github.com/duckdb/duckdb-httpfs/pull/165),
+still unmerged). Until that lands, treat `use-credential-chain` as suitable for short-lived queries
+and for workloads that tolerate connection recycling, and expect failures wherever a single
+connection must serve reads for longer than the credential lifetime.
```
ferenc-csaky, line 52: Can we make this a bit more general and hopefully more concise? IMO this is way too much implementation detail for a general user facing doc. For example referencing that DuckDB PR is completely unnecessary.

ferenc-csaky (commented):
Logic LGTM, added 1 comment about the docs.

PR #2099, opened 2026-05-27
fix: skip _all default partition for RANGE+TTL tables managed by pg_partman

WellMafra:
Companion to [DataSQRL/cloud-compilation#939](https://github.com/DataSQRL/cloud-compilation/pull/939).

`PostgresCreateTableDdlFactory` skips the `_all` catch-all partition for RANGE+TTL tables — pg_partman owns those. HASH tables and RANGE tables without TTL are unchanged.

Snapshot updated: `OrderUpdates_1` (RANGE, 30d TTL) no longer emits the `_all` DDL.


ferenc-csaky:
Superseded by https://github.com/DataSQRL/sqrl/pull/2147

PR #2118, opened 2026-06-05
feat: Resolve available, non-secret environment variables during `compile`

ferenc-csaky:
## Key Changes
- Add `SecretEnvVarResolver` that will resolve available regular env var definitions, and defines a new secret env var template `${{...}}` that will be resolved to `${...}` in `compile
- `compile` env var resolution is applied for
  - The table properties of any Flink `CREATE TABLE` statements
  - Leaf nodes under the `connectors` key in any user-provided `package.json`

More detailed behavior explanation is added to the docs.

## TODO
- [x] Requires a change in the runner first
- [x] Also resolve connector config options
- [x] Update docs and explain secret env vars

### DIFF
--- a/documentation/docs/configuration-default.md
+++ b/documentation/docs/configuration-default.md
@@ -108,10 +108,7 @@ The connector templates configured under `connectors` can use environment variab
 
 ### Environment Variables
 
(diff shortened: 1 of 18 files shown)


mbroecheler (approved):
The structure and setup make sense to me. Nicely done!
Do we have test coverage that an env var which is not available at compile time is left in place?

PR #2124, opened 2026-06-08
fix: Make sure Vert.x BOM does not override Jackson version

ferenc-csaky:


sqrl-cli/pom.xml
```diff
@@ -32,6 +32,12 @@
   </properties>
 
   <dependencies>
+    <!-- Required to precede flink-sql-runner, which shades flink-connector-http that ships an old version. -->
```
velo, line 35: don't you love when people only publish shaded jars?!

velo (approved):

sqrl-cli/pom.xml
ferenc-csaky: if they not, that's also a problem 😄

PR #2127, opened 2026-06-10
feat: Enable compression for Vert.x responses

ferenc-csaky:
## Key Changes
- Enable compression support for Vert.x server responses in case the client accepts it
  - Compress size limit is set to 25MB for now 
- Refactor `server-config.json`, eliminate duplicate keys


mbroecheler (commented):
Great, so it was already present in the configs, we just needed to wire it up. Nice.

Is there a way to test this? Can we enable compression in our test harness when we execute the graphql queries and mutations for example (if server supports it) to make sure this works?

PR #2128, opened 2026-06-10
feat: Enable compression by default for Kafka producers

ferenc-csaky:


sqrl-planner/src/main/resources/default-package.json
```diff
@@ -47,6 +47,7 @@
       "properties.bootstrap.servers": "${KAFKA_BOOTSTRAP_SERVERS}",
       "properties.group.id": "${KAFKA_GROUP_ID}",
       "properties.auto.offset.reset": "earliest",
+      "properties.compression.type": "zstd",
       "topic": "${sqrl:table-name}"
     },
     "kafka": {
```
mbroecheler, line 53: Why not add compression for the subscription topics as well?

ferenc-csaky: My first thought was it's irrelevant for consumers, and then Kafka will complain about it, that an unused prop is supplied, but 1) that's already happening for other configs, and it's an INFO level log, and 2) there is no sane way to detect if a Flink SQL kafka table will be used as consumer or a producer. So added it. :)

PR #2129, opened 2026-06-11
feat: Stop Flink job with drain in test executions

ferenc-csaky:
## Key Changes
- Instead of Flink `cancel`, use stop with advancing watermark to the end of event time to make sure no event is stuck in the pipeline
- Since the Flink API only exposes `stopWithSavepoint`, we cannot omit taking a savepoint, which gets saved to a temp dir, and then deleted right away
- This stop action will now trigger a checkpoint failure (expected, cause the job is already stopping), added some logic to exclude those warnings with stacktraces from the output
- Removed the now unnecessary watermark advance record from the `banking` test data, and updated every snapshot that used it


mbroecheler:
@ferenc-csaky I am surprised that we don't have more tests that are affected by this. The only one seems to be banking, right?

ferenc-csaky:
Even that was not affected, cause it ran before I changed the execution flow to only execute the test queries after the Flink job exited.

The changes were necessary cause the record i removed affected the calculations (which points to the same direction, it was already encountered before this change).

mbroecheler:
@ferenc-csaky I would suggest we add a time window query to our test cases - those should definitely be affected because the last window should not have closed before this change.

PR #2131, opened 2026-06-12
feat: Introduce `test(no_rows)` to skip snapshot generation for empty test queries

### LINKED ISSUE #2130: Introduce `test(no_rows)` hint
Currently `/*+test */` creates a snapshot for the results returned by the table the hint is annotating. This works great for tests where users want to assert a certain result set.

However, we are increasingly seeing users writing tests to assert that certain invariants are true and the way those are written is that they return empty tables if they are green (i.e. no records violate the invariant/condition) or the violating records if there are test failures.

That works great for testing conditions on lots of test data, but its a bit awkward with the current snapshot testing because it creates empty snapshots for all these tests and those need to be checked into the repository.

It would be cleaner to add a `/*+test(no_rows) */` hint which extends the current hint with a parameter: `no_rows`. When that is set, the assertion is that the table returns no records. Hence, we don't need to create a snapshot: the test runner validates that the result set is empty and if it is not, it prints out all records as test failures.

ferenc-csaky:
## Key Changes
- Add `no_rows` arg to the `test` hint
- Add a `TestType` enum to every `GraphqlQuery` object in the test plan
- Fetch `TestType` from each test table based on the hnt value
- Add new snapshot result types print format for `no_rows` tests
- Evaluate `no_rows` tests in `DatasqrlTest` properly
- Introduced a `no_rows` test to an integration test case

### DIFF
--- a/sqrl-cli/src/main/java/com/datasqrl/compile/TestPlan.java
+++ b/sqrl-cli/src/main/java/com/datasqrl/compile/TestPlan.java
@@ -18,7 +18,6 @@ package com.datasqrl.compile;
 import com.datasqrl.engine.database.relational.JdbcStatement;
 import java.util.List;
 import java.util.Map;
-import javax.annotation.Nullable;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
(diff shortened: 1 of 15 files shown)


sqrl-cli/src/main/java/com/datasqrl/compile/TestPlan.java
```diff
@@ -36,9 +37,16 @@ public class TestPlan {
   @NoArgsConstructor
   @Getter
   public static class GraphqlQuery {
+
+    public enum TestType {
+      REGULAR,
```
mbroecheler: Let's call this SNAPSHOT - `REGULAR` is a bit generic ;-P

PR #2132, opened 2026-06-12
feat: Add support for visualizing data model in Voyager page

VolodymyrMintenko:


sqrl-cli/src/main/java/com/datasqrl/compile/CompilationProcess.java
```diff
@@ -90,6 +90,12 @@ public Pair<PhysicalPlan, TestPlan> executeCompilation(Optional<Path> testsPath)
 
       loadResult.inferredSchema().ifPresent(writeDeploymentArtifactsHook::writeInferredSchema);
 
+      // Data model (GraphQL schema) Voyager page; first API version when multi-version.
+      if (!apiVersions.isEmpty()) {
+        writeDeploymentArtifactsHook.writeDataModelVisual(
+            apiVersions.get(0).schema().getDefinition());
```
ferenc-csaky, line 96: Pick that latest elem of the list instead of the first. I'll make this list sorted in an upcoming change, but the idea here is if we have multiple API versions, always pick the latest.

ferenc-csaky (changes requested):

PR #2134, opened 2026-06-12
chore: Make `ApiSources` comparable and sort them

ferenc-csaky:


sqrl-planner/src/main/java/com/datasqrl/graphql/ApiSources.java
```diff
@@ -17,11 +17,17 @@
 
 import java.util.List;
 
-public record ApiSources(String version, ApiSource schema, List<ApiSource> operations) {
+public record ApiSources(String version, ApiSource schema, List<ApiSource> operations)
```
velo, line 20: don't you want to `@NonNull String version`?! So lombok protects you

velo (approved):

sqrl-cli/src/main/java/com/datasqrl/compile/CompilationProcess.java
```diff
@@ -86,12 +88,13 @@ public Pair<PhysicalPlan, TestPlan> executeCompilation(Optional<Path> testsPath)
           serverPlan.getMutations().size());
 
       var loadResult = graphqlSourceLoader.load(serverPlan);
-      var apiVersions = loadResult.apiVersions();
+      var apiVersions = new ArrayList<>(loadResult.apiVersions());
```
velo, line 91: var apiVersions = loadResult.apiVersions().stream().sorted().toList();

sqrl-planner/src/main/java/com/datasqrl/graphql/ApiSources.java
ferenc-csaky: protecting me from what, failing with an NPE 1 step earlier? :) but that version cannot be `null` though, we either set it from a static const, or from the JSON config, which is validated

sqrl-cli/src/main/java/com/datasqrl/compile/CompilationProcess.java
ferenc-csaky: more chars, and imo less readable too

PR #2161, opened 2026-06-23
feat: Support shared scripts

ferenc-csaky:
## Key Changes
- Rename mounted dir from `build` to `workspace` inside Docker container
- Introduce new optional CLI option `-r, --project-root` to explicitly set relative project root to the mount point
- If project root is not set, it gets inferred from the given package file(s), and it will be the deepest common subdir (e.g. `a/b/c/d/package.json`, `a/b/c/package.json`, `a/b/package.json` ->`a/b`)
- Introduce `shared` config option under `main` (see details below)
- Add `root` prefix for import paths (`IMPORT root.my.import.script;`), which will always try to resolve the given path against the project root folder, even if the defining script is in a submodule
- Moved over SQRL script processing to a new `SqrlPreprocessor`, which now deals with the shared scripts, and the Mustache template resolution

### Shared script concept
We can now setup packages that we plan to share between multiple projects. Most probably these will be data catalogs. It can be arranged as regular SQRL project, with the restriction the it requires a `package.json` named config in its root:
```
banking-shared
├── application_updates.jsonl
├── applications.jsonl
├── customers.jsonl
├── loan_types.jsonl
├── package.json
└── sources.sqrl
```
That package JSON can be minimal, and may contain an information about the project and define config variables if the defined shared script content has any of them:
```json
{
  "version": "1",
  "information": {
    "description": "Common table definitions for banking test use cases"
  },
  "script": {
    "config": {
      "monitorInterval": "10s"
    }
  }
}
```
Then, we can add `banking-shared` to our regular SQRL project's package like:
```json5
{
...
"script": {
  "main": "loan.sqrl",
  "shared": {
    "banking-catalog": { // name for the imported script, will be used when referenced in SQRL
      "path": "../banking-shared" // path for the shared script's dir
      "config": { "monitorInterval": "5s" } // optional template variable overrides
    }
  },
  ...
}
```
The key name under `shared` will define the name of the shared module, which it can be imported by.
A shared script `IMPORT` will look like this:
```sql
# In loan.sqrl
IMPORT `banking-catalog`.sources;

# OR
IMPORT root.`banking-catalog`.sources;
```
> [!IMPORTANT]
> In most cases the `root` prefix is optional, except when the caller script is in a subdirectory, it has to be present to be able to import from the build root.

## TODO
- [x] Fix any failing tests
- [x] Add test case for shared script template config and shared template config override
- [x] Update docs and README


sqrl-testing/sqrl-testing-integration/src/test/java/com/datasqrl/FullUseCaseRunner.java
```diff
@@ -75,7 +74,7 @@ public static void run(UseCaseParam param, TestContainerHook containerHook) {
       throw e;
     }
 
-    var rootDir = param.packageJsonPath().getParent();
+    var preojectRoot = param.packageJsonPath().getParent();
```
mateczagany: typo: `projectRoot`

sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/SqrlPreprocessor.java
```diff
@@ -0,0 +1,108 @@
...
+    } catch (Exception e) {
+      throw new IllegalStateException("Failed to resolve template in SQRL file: " + file, e);
+    }
+  }
+
+  private Map<String, Object> createTemplateValues(Path scriptDir) {
+    var sharedScripts = config.getScriptConfig().getSharedScriptConfigs();
+
+    for (var sharedScript : sharedScripts) {
+      if (scriptDir.endsWith(sharedScript.getPath())) {
```
mateczagany: This would fail to match the correct configuration for nested scripts, such as `banking-shared/subdir/sources.sqrl` and fallback to the project config

sqrl-planner/src/main/resources/jsonSchema/packageSchema.json
```diff
@@ -146,6 +146,28 @@
...
+        },
+        "shared": {
+          "type": "object",
+          "minProperties": 1,
+          "patternProperties": {
+            "^[A-Za-z0-9_-]+$": {
+              "type": "object",
+              "minProperties": 1,
+              "properties": {
+                "path": {
```
mateczagany, line 158: This should be made required.

sqrl-cli/src/main/java/com/datasqrl/packager/Packager.java
```diff
@@ -59,22 +53,29 @@
...
-  private final FilePreprocessingPipeline preprocPipeline;
+  private final FilePreprocessingPipeline preprocessingPipeline;
   private final MainScript mainScript;
 
   public void preprocess(ErrorCollector errors) {
-    errors.checkFatal(
-        config.getScriptConfig().getMainScript().map(StringUtils::isNotBlank).orElse(Boolean.FALSE),
-        "No config or main script specified");
+    // Validate main script
+    config.getScriptConfig().getMainScript();
```
mateczagany, line 63: Why remove the validation here?

sqrl-cli/entrypoint.sh
```diff
@@ -22,14 +22,20 @@ if [[ -n "${SQRL_DEBUG+x}" && -n "$SQRL_DEBUG" ]]; then
   set -x
 fi
 
-cd /build
+if [[ -d /build ]]; then
+  WORKSPACE=/build
+  printf 'Warning: The "/build" mounting point is deprecated and will be removed, please use "/workspace" instead\n' >&2
+elif [[ ! -d /workspace ]]; then
+  printf 'Error: The "/workspace" mount does not exist, try adding "-v $PWD:/workspace" to the docker command' >&2
```
mateczagany: Add trailing `\n` to the error as well

sqrl-planner/src/main/java/com/datasqrl/plan/MainScriptImpl.java
```diff
@@ -32,32 +32,30 @@
...
-    return config
-        .getScriptConfig()
-        .getMainScript()
-        .map(Path::of)
-        .flatMap(resourceResolver::resolveFile);
+    var mainScript = config.getScriptConfig().getMainScript();
+    var mainScriptPath =
+        workspacePaths.projectRoot().map(p -> p.resolve(mainScript)).orElse(Path.of(mainScript));
+
+    return resourceResolver.resolveFile(mainScriptPath);
```
mateczagany, line 54: The `mainScriptPath` will be an absolute path after this change, and `resourceResolver` will resolve the unprocessed source file, not the one under the `build` dir which would be the baseDir of `resourceResolver`.

```
❯ docker run --rm \
  -v "$PWD/sqrl-testing/sqrl-testing-integration/src/test/resources/usecases:/workspace" \
  datasqrl/cmd:local compile -r avro-schema

...

[FATAL] Object [_Schema] not found in module: _Schema
in script:avro-schema.sqrl [1:1]:
IMPORT avro-schema.schema_{{variant}} AS _Schema;
```

sqrl-cli/src/main/java/com/datasqrl/packager/Packager.java
ferenc-csaky: Calling `getMainScript()` will do that validation, so i eliminated the more complex code

sqrl-planner/src/main/java/com/datasqrl/plan/MainScriptImpl.java
ferenc-csaky: This is a very good catch, this should have been reverted to how it was, and I forgot about it.

sqrl-cli/src/main/java/com/datasqrl/cli/BaseOsProcessManagerCmd.java
```diff
@@ -16,22 +16,39 @@
...
+      description =
+          "Project root folder. Must be a relative path. If omitted, it is inferred from the package path(s), falling back to \"/workspace\".")
+  protected Optional<Path> projectRoot = Optional.empty();
 
   @Option(
       names = {"-t", "--target"},
-      description = "Target folder for deployment artifacts and plans. Default: \"build/deploy\".")
-  protected Path targetFolder = DEFAULT_TARGET;
+      description =
+          "Target folder for deployment artifacts and plans. Must be a relative path.%nDefault: \"<project-root>/build/deploy\".")
```
mbroecheler: relative to mount dir or project root?

ferenc-csaky: if it is set explicitly, relative to mount dir, it ignores project root now.

mateczagany (approved):

sqrl-cli/src/main/java/com/datasqrl/cli/BaseOsProcessManagerCmd.java
ferenc-csaky: After giving it a thought, probably the most straightforward thing to do would be to in case the `project-root` is explicitly set, the `targetFolder` should be relative to that path.

Originally I left it relative to mount dir, cause that seemed the obvious thing, but once someone types in the package files relative to an explicitly set project root, doing the same with the `targetFolder` is what I would expect as a user.

PR #2165, opened 2026-06-25
docs: document dedicated-nodes and do-not-disrupt deployment options

velo:
Documents two cloud-deployment options in `documentation/docs/configuration-engine/cloud-deployment.md`:

- **`*-node-groups`** — pins a component's pods onto dedicated nodes (hard requirement → Pending if unavailable). Each name derives a node selector (`N=true`) + a toleration for taint key `N`; infra labels+taints the node group. Covers `taskmanager-node-groups`/`jobmanager-node-groups` (Flink), `node-groups` (Postgres, Vert.x).
- **`do-not-disrupt`** — protects pods from voluntary autoscaler consolidation. Documents per-engine defaults (Flink false, Postgres true, Vert.x false).

Matches the implementation in cloud-compilation (per-service deployment config passthrough).

### DIFF
--- a/documentation/docs/configuration-engine/cloud-deployment.md
+++ b/documentation/docs/configuration-engine/cloud-deployment.md
@@ -116,7 +116,8 @@ PostgreSQL deployments consist of one primary instance and a configurable number
         "replica-count": 1,             // Number of read replicas (0 or larger)
         "disk-size-gb": 256,            // Disk size in GB (1 or larger)
         "auto-expand-percentage": 0.2,  // Auto-expand threshold (0 to disable, must be < 1)
-        "create-indexes": true          // Whether to create table indexes (see "Create Indexes" below)
+        "create-indexes": true,         // Whether to create table indexes (see "Create Indexes" below)
+        "parameters": {}                // Extra postgresql.parameters (see "Parameters" below)
       }
     }
   }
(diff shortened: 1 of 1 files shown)


velo:
Related PRs — ship together:
- DataSQRL/cloud-compilation#1030 — `create-indexes` option, schema index-strip, and the `CreatePostgresIndexes` upgrade step
- DataSQRL/cloud-backend#553 — metrics catch-up skips indexes; blue-green validation moved after the steady-state upgrade
- DataSQRL/sqrl#2165 — `create-indexes` documentation in `cloud-deployment.md`

ferenc-csaky:
I'm wondering about if it's a good thing in general to include cloud-specific things to the doc? Technically these options are orthogonal to SQRL, at this point cloud processes and validates them.

I guess it makes sense to keep documentation centralized, but then I think it would make sense to restructure the docs and separate Cloud-specific things to their own page or something.

documentation/docs/configuration-engine/cloud-deployment.md
```diff
@@ -162,4 +164,143 @@ Vert.x API server deployments consist of a configurable number of identically si
...
+    }
+  }
+}
+```
+
+---
+
+## Create Indexes (`create-indexes`)
+
+Controls whether the PostgreSQL table indexes are created for the deployment. Defaults to `true`. PostgreSQL only.
```
mbroecheler, line 261: it only applies to secondary indexes, right? Not the primary key indexes that are defined on the table itself. Those are needed even during replay, otherwise Postgres cannot effectively upsert and every upsert becomes a full table scan.

velo: Correct, we only strip the `CREATE INDEX` statements

https://github.com/DataSQRL/cloud-compilation/pull/1030/changes#diff-aa3842a78f418d9475334340a2b1662303de024c9d880a112f8e7d95f4a5978aR84-R92


no PKs are hurt on this process

ferenc-csaky (approved):

PR #2169, opened 2026-06-26
fix: Decode gzip request bodies and enable gzip responses by default

### LINKED ISSUE #2168: Gzip-compressed request body to REST mutation endpoint fails with misleading 'JWT auth failed'
## Summary

POSTing a gzip-compressed body (`Content-Encoding: gzip`) to a REST mutation endpoint fails with a misleading **`JWT auth failed`** error. The auth/JSON layer parses the **raw, still-compressed** request body, choking on the gzip magic byte before the body is ever decompressed. The JWT itself is valid.

## Reproduction

```bash
echo '{
  "event": [
    {
      "deploymentId": "dd832b18-8183-4303-b52d-54e7d0a1d2cb",
      "instanceId": "flink-application",
      "log": "test log",
      "service": "flink"
    }
  ]
}' | gzip | curl -X POST "https://metrics2.api.cloud.sqrl.site/v1/rest/mutations/DeploymentLogs" \
  -H "Content-Type: application/json" \
  -H "Content-Encoding: gzip" \
  -H "Authorization: $JWT_TOKEN" \
  --data-binary @-
```

### Actual (HTTP 500)

```json
{"error":"JWT auth failed","cause":"JsonParseException: Illegal character ((CTRL-CHAR, code 31)): only regular white space (\\r, \\n, \\t) is allowed between tokens\n at [Source: REDACTED ...); line: 1, column: 2]"}
```

`CTRL-CHAR code 31` = `0x1F`, the first byte of the gzip magic number (`0x1f 0x8b`). The request body is being JSON-parsed while still gzip-compressed.

The same request **without** gzip (no `| gzip`, no `Content-Encoding: gzip` header) returns HTTP 200:

```json
{"data":[{"timestamp":"2026-06-26T14:29:47.108Z"}]}
```

## Impact

- Clients cannot compress request bodies, which matters for large log/metric batches to endpoints like `DeploymentLogs`.
- The error message (`JWT auth failed`) is misleading and sends users debugging their tokens instead of the request encoding.

## Suggested fix

Decode `Content-Encoding: gzip` on the request body **before** the auth filter / JSON parsing runs (filter ordering), or explicitly reject the header with a clear error if gzip request bodies are intentionally unsupported.

velo:
## Problem

POSTing a gzip-compressed body (`Content-Encoding: gzip`) to any POST endpoint (`/v1/graphql`, REST mutations `/v1/rest/mutations/*`, MCP) failed. The body reached the JSON parser **still compressed**, choking on the gzip magic bytes (`0x1f 0x8b` → `CTRL-CHAR code 31`). On JWT-protected routes the `JwtFailureHandler` then masked the parse error as a misleading **`JWT auth failed`**, sending users to debug valid tokens. Fixes #2168.

Root cause: Vert.x `HttpServerOptions` shipped with `decompressionSupported=false`, so request bodies were never decompressed. Separately, response compression — though `compressionSupported=true` — never fired because `compressionContentSizeThreshold` was 25 MB.

## Fix

Two lines in the shipped default config template (`sqrl-cli/.../templates/server-config.json`), the single source for the deployed `vertx-config.json`, so it covers REST, GraphQL, and MCP at once with no code change:

- `decompressionSupported: false → true` — decode gzip request bodies before any handler runs
~- `compressionContentSizeThreshold: 26214400 → 0` — let responses gzip-compress by default when the client sends `Accept-Encoding: gzip`~

## Test

New Docker container IT `GzipContainerIT` covers gzip **both ways** × **REST mutation + GraphQL** × **with/without JWT** (two deployments: `sensors-mutation` no-auth, `jwt-authorized` with JWT). Built TDD-style — confirmed it fails on the unfixed build first:

| | gzip request (GraphQL) | gzip request (REST mutation) | gzip response |
|---|---|---|---|
| pre-fix | `400` ❌ | `500` ❌ | uncompressed ❌ |
| post-fix | `200` ✅ | `200` ✅ | `Content-Encoding: gzip` ✅ |

`mvn test` green; `GzipContainerIT` green via `mvn -B install -DonlyContainerE2E -Dit.test=GzipContainerIT`.

## Out of scope

Rewording the `JwtFailureHandler` message — the misleading error disappears once the body parses.


sqrl-cli/src/main/resources/templates/server-config.json
```diff
@@ -27,13 +27,13 @@
     "activityLogDataFormat" : "HEX_DUMP",
     "alpnVersions" : [ "HTTP_2", "HTTP_1_1" ],
     "clientAuth" : "NONE",
-    "compressionContentSizeThreshold" : 26214400,
```
ferenc-csaky, line 30: I'd keep this, 25MB should be enough for everything.

ferenc-csaky (changes requested):

ferenc-csaky (approved):

sqrl-cli/src/main/resources/templates/server-config.json
velo: That is like the minimal threshold for compression to kick in.... I think half to 1 MB would be a territory where compression makes sense

PR #2170, opened 2026-06-29
fix: Remap dynamic params inside scalar subqueries

mateczagany:
Without this fix, dynamic params inside scalar subqueries would not get remapped properly, causing invalid/unexpected queries. 

- Traverse `RexSubQuery.rel` when replacing SQRL function dynamic parameters
- Add a DAG planner test for this exact issue


ferenc-csaky (approved):
Thanks for fixing this!

PR #2178, opened 2026-06-29
fix: Preserve applied `IMPORT` hints on importing another `.sqrl` script

### LINKED ISSUE #2102: Hints on IMPORT (ttl, no_query, insert) are silently ignored instead of applied or rejected
## Summary

The `ttl(...)` hint behaves differently depending on whether it is placed on a `CREATE TABLE` definition or on an `IMPORT` statement that imports that table. On `CREATE TABLE` it works; on `IMPORT` it is **silently ignored** — no error, no warning, no effect on the compiled plan. This makes it easy to write a script that looks correct but produces no retention.

Observed with `datasqrl/cmd:0.10.5`.

## Expected

One of the following, consistently:
- The `ttl` hint on `IMPORT` is honored (propagated to the imported table's engine config), **or**
- The compiler emits a warning/error that `ttl` (and other definition-level hints) are not valid on an `IMPORT` statement.

Today it does neither — the hint is accepted by the parser and then dropped.

## Actual

`ttl` on `IMPORT` produces no `retention.ms` in `kafka.json` (topic config stays `{}`), and `ttl: 0` everywhere in `postgres.json` / `flink-compiled-plan.json`. Moving the identical hint onto the `CREATE TABLE` definition produces the expected `retention.ms`.

## Trivial reproduction

`mytable/MyEvents.sqrl` — the table definition:

```sql
/*+ no_query, engine(kafka) */
CREATE TABLE MyEvents (
    `id` STRING NOT NULL METADATA FROM 'uuid',
    `payload` STRING,
    `timestamp` TIMESTAMP_LTZ(3) NOT NULL METADATA FROM 'timestamp',
    WATERMARK FOR `timestamp` AS `timestamp` - INTERVAL '0.001' SECOND
);
```

`main.sqrl` — case A, hint on the IMPORT (ignored):

```sql
/*+ no_query, ttl(14 days) */
IMPORT mytable.MyEvents.*;
```

Compile → `kafka.json`:

```json
{ "topicName": "MyEvents", "config": {} }   // <-- ttl dropped, no retention.ms
```

`main.sqrl` — case B, hint moved onto the CREATE TABLE (works):

```sql
/*+ no_query, engine(kafka), ttl(14 days) */
CREATE TABLE MyEvents ( ... );
```
```sql
-- main.sqrl
IMPORT mytable.MyEvents.*;
```

Compile → `kafka.json`:

```json
{ "topicName": "MyEvents", "config": { "retention.ms": "1209600000" } }   // <-- 14 days
```

## Impact

A user reasonably expects to set per-import retention from the consuming script (the table definition may be a shared/imported package they don't own). Because the hint is silently dropped, topics silently fall back to whatever engine-level default retention is configured (e.g. a global `kafka.retention`), which can be effectively "keep forever" — a real cost/correctness footgun with no diagnostic.

## Related observation (separate, lower priority)

While investigating, `no_query` and `insert(batch)` placed on an `IMPORT` of a kafka source table also had no effect (stripping them yielded a byte-identical plan). If definition-level hints are generally not meant to apply on `IMPORT`, a single validation pass that rejects/warns on unsupported hints-on-import would cover all of these cases.

ferenc-csaky:
## Key Changes
- Manage inherited imports on `planMain`
- Apply inherited imports if there are any when we create `PlannerHints` during SQRL script planning
- Added DAG planner test case to cover the inherit logic

Fixes #2102

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/planner/SqlScriptPlanner.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/planner/SqlScriptPlanner.java
@@ -211,17 +211,19 @@ public class SqlScriptPlanner {
   }
 
   /**
-   * Main entry method for parsing a SQRL script. The bulk of this method ensure that exceptions and
-   * errors are correctly mapped to the source so that users can easily understand what the issue is
-   * and what's causing it.
+   * Main entry method for parsing a SQRL script. The bulk of this method ensures that exceptions
+   * and errors are correctly mapped to the source so that users can easily understand what the
+   * issue is and what's causing it.
    *
-   * @param mainScript
-   * @param sqrlEnv
+   * @param mainScript SQRL script to plan
+   * @param inheritedHints inherited hints if the planned script is coming from an IMPORT
+   * @param sqrlEnv the SQRL compilation environment
    */
   public void planMain(
       MainScript mainScript,
       Optional<PlannerHints> inheritedHints,
       Sqrl2FlinkSQLTranslator sqrlEnv) {
+
     var scriptErrors = errorCollector.withScript(mainScript.getPath(), mainScript.getContent());
     var statements = sqrlParser.parseScript(mainScript.getContent(), scriptErrors);
     var statementStack = new ArrayList<StackableStatement>();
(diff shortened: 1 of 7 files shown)


sqrl-planner/src/main/java/com/datasqrl/planner/SqlScriptPlanner.java
```diff
@@ -218,7 +218,10 @@ public SqlScriptPlanner(
    * @param mainScript
    * @param sqrlEnv
    */
-  public void planMain(MainScript mainScript, Sqrl2FlinkSQLTranslator sqrlEnv) {
+  public void planMain(
```
mateczagany, line 222: The JavaDocs above was not updated

mateczagany (approved):

PR #2191, opened 2026-07-06
feat: Add opt-in OffsetPageInfo metadata for paginated GraphQL queries

velo:
## What

Adds **opt-in pagination metadata** for user-authored GraphQL schemas. A query whose result type is a *page wrapper* — an object with a list field plus a `OffsetPageInfo` field — returns its rows plus pagination metadata computed from a companion `COUNT(*)` query.

```graphql
type Query {
  activityLogs(fromTime: DateTime!, toTime: DateTime!, limit: Int = 100, offset: Int = 0): ActivityLogEntryPage!
}
type ActivityLogEntryPage {
  results: [ActivityLogEntry!]
  pagination: OffsetPageInfo
}
```

The standard `OffsetPageInfo` type is auto-injected when referenced but not defined (and validated for an exact match when the user defines it):

```graphql
type OffsetPageInfo {
  totalRecords: Long!
  pageSize: Int!
  currentPage: Int!
  totalPages: Int!
  hasNextPage: Boolean!
  hasPreviousPage: Boolean!
  nextOffset: Int          # null when no next page; feeds straight back into `offset`
  prevOffset: Int          # null on first page
  firstEventTime: DateTime # MIN(rowtime col); null if the result has no rowtime column
  lastEventTime: DateTime  # MAX(rowtime col)
}
```

## How

- **Opt-in.** Existing list queries are untouched. For inferred/no-schema mode, the new compiler setting `compiler.api.paginated-results` (default `false`) generates `<Type>Page {results, pagination}` wrappers plus the `OffsetPageInfo` type for every multi-row query and relationship; with the default the inferred schema is unchanged.
- **Detection is shape-based** (flexible field names): an object with exactly two fields — one a list of an object type, one of type `OffsetPageInfo`. Only valid for `MANY` query functions that declare `limit`/`offset`.
- **Compiler** (`sqrl-planner`): `GraphqlSchemaWalker` sees through the wrapper and validates/walks the element type against the table-function row type; `GraphqlModelGenerator` emits two companion aggregate queries over `FROM (<baseSql>) x`: `countSql` (`COUNT(*)` only) and `countWithEventTimesSql` (`COUNT(*)` + `MIN/MAX` over the rowtime column, absent when the result has no rowtime). `OffsetPageInfoUtil` injects/validates the standard type.
- **Runtime** (`sqrl-server`): `SqlQuery` gains nullable `countSql`/`countWithEventTimesSql` (`@JsonInclude(NON_NULL)` → old `vertx.json` still loads, existing snapshots unchanged). `VertxQueryExecutionContext` computes metadata **lazily from the selection set**: the aggregate query only runs when `totalRecords`/`totalPages` (count variant) or `firstEventTime`/`lastEventTime` (count+MIN/MAX variant) are selected, and runs in parallel with the data query. When only `hasNextPage`/`nextOffset` are selected, no aggregate runs — the data query fetches `LIMIT+1` rows and the extra row is trimmed. Selecting only `results` (or only offset-derived fields like `currentPage`/`hasPreviousPage`) adds no extra work at all. Metadata math is a pure static method.

## Tests

- **`PaginationMetadataTest`** — 7 unit cases for the metadata math (empty, first/middle/last/beyond-end pages, `limit=0`, event-time passthrough).
- **`UseCaseCompileTest`/`DAGWriterJsonTest`** — new `clickstream/package-paginated.json` usecase snapshots the generated paginated schema (wrappers, `OffsetPageInfo`, count SQL with bound parameters); all other usecase snapshots are byte-identical, proving the default is backwards compatible.
- **`PagedQueryIT`** — real-Postgres test with a recording `SqlClient` proving the lazy execution: no aggregate query when only `results` or offset-derived fields are selected, LIMIT+1 (and trimming) for `hasNextPage` alone, plain-COUNT variant for totals, COUNT+MIN/MAX variant for event times.
- **`GraphQLValidationTest`** — 6 new schema cases: 2 positive (auto-injected type; user-defined type with non-conventional field names + a paged relationship) and 4 negative (wrong pagination type, missing limit/offset, paged subscription, unknown element field).
- All touched-module unit tests pass; the 3 aggregate validation snapshots changed only because they enumerate every input `.graphqls`.

## Follow-up

Runtime execution (count query + wrapper assembly) is exercised by unit tests and consistent with the existing list path, but a **full-pipeline usecase under `FullUseCaseIT`** (Postgres/Flink/Kafka containers) asserting `OffsetPageInfo` payloads across page boundaries is a recommended follow-up — it needs the heavy container harness.


mbroecheler:
I like the idea to add this. Would be nice if we could also generate this GraphQL schema code.

Couple of thoughts:
1. I would prefer a neutral name like `OffsetPageInfo` instead of SqrlPagination. Since this needs to be hardcoded, we don't want `sqrl` to be leaking through user's API schema.
2. A critical aspect that OffsetPageInfo is computed lazily based on what users are actually selecting in their queries. In particular `first/lastEventTime` is expensive to compute and should only be retrieved when it is actually needed. Same for `hasNextPage` - which I assume is computed by fetching `LIMIT +1` - we don't need to fetch that extra record if that field isn't queried. Same goes for totalRecords / totalPages.

velo:
>  Would be nice if we could also generate this GraphQL schema code.

I think we would need a hint for that right?!

Or do we want to always generate as paginated?

mbroecheler:
> I think we would need a hint for that right?!
> 
> Or do we want to always generate as paginated?

Yes, but not a hint - it would be a compiler config setting to generated paginated, default false (for now to be backwards compatible).

velo:
Neutral name `OffsetPageInfo` :heavy_check_mark: 
Lazy on `OffsetPageInfo` fields :heavy_check_mark: 
Configuration to enable `OffsetPageInfo` :heavy_check_mark:

sqrl-planner/src/main/java/com/datasqrl/config/GraphqlSourceLoader.java
```diff
@@ -75,9 +76,17 @@ public LoadResult load(ServerPhysicalPlan serverPlan) {
...
-      apiVersions.forEach(
-          apiVersion -> graphqlSchemaHandler.validateSchema(apiVersion, serverPlan));
-      return new LoadResult(apiVersions, Optional.empty());
+      var injected =
+          apiVersions.stream()
+              .map(
+                  apiVersion ->
+                      new ApiSources(
+                          apiVersion.version(),
+                          OffsetPageInfoUtil.injectPaginationType(apiVersion.schema()),
```
ferenc-csaky: Do we only want to do this for of the user-provided schema and also regardless of the `paginated-results` config value?

velo: Yes, this would be injected only if user missed it.  But now, I think this is wrong.
Going to change to an all or nothing proposition.  Either you declare the OffsetPageInfo, so you can use it, or you can't use it at all.  This will also create a self containing graphql schema.

Changed to validate only, but not inject it.

sqrl-planner/src/main/java/com/datasqrl/server/GraphqlModelGenerator.java
```diff
@@ -186,6 +192,34 @@ protected void visitQuery(
...
+            col ->
+                "SELECT COUNT(*) AS \"total_records\", MIN(\""
+                    + col
+                    + "\") AS \"first_event_time\", MAX(\""
+                    + col
+                    + "\") AS \"last_event_time\" FROM ("
+                    + baseSql
+                    + ") x")
+        .orElse(null);
+  }
```
ferenc-csaky: I do not think this is viable. Executing the whole query to get the maximum results seems like a big nono. Also, in a streaming environment `total_records` is an undefined thing. Well it can be something at T timepoint, but most probably it will be something else in T+1

sqrl-planner/src/main/java/com/datasqrl/server/GraphqlModelGenerator.java
```diff
@@ -168,13 +169,18 @@ protected void visitQuery(
             .map(InputValueDefinition::getName)
             .anyMatch(
                 name -> name.equals(SchemaConstants.LIMIT) || name.equals(SchemaConstants.OFFSET));
+    var countSql = paged ? buildCountSql(executableJdbcReadQuery.getSql()) : null;
```
ferenc-csaky: We should introduce a new `PaginationType` (I guess `OFFSET_PAGE_INFO`), and set that here for the `SqlQuery`.

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/server/jdbc/VertxQueryExecutionContext.java
ferenc-csaky, line 1: These changes in this class are write-only AI code completely and i do not like it one single bit.

The newly introduced `OffsetPageInfo` reuses the `LIMIT_AND_OFFSET` code and it results in some weird mixed execution I do not want to debug 3months from now.

I suggested to introduce a new pagination type in my other comment. We should use that here and separate the executions, possibly reusing any common logic needed for `LIMIT_AND_OFFSET`, but we need clear abstraction layers and an execution flow that a human can debug without losing their sanity.

ferenc-csaky (changes requested):

sqrl-planner/src/main/java/com/datasqrl/server/GraphqlModelGenerator.java
velo: Good point.  For some queries the total_records/total_pages might be just fine.  For other it might be really slow to fetch them.

I will drop both for now, let's get feature out of the door, and if other people besides us point out how that is necessary we can work on it later.

sqrl-planner/src/main/java/com/datasqrl/server/GraphqlModelGenerator.java
```diff
@@ -186,6 +207,37 @@ protected void visitQuery(
     queryCoords.add(coordsBuilder.build());
   }
 
+  /** Select expression behind {@code totalRecords}/{@code totalPages}. */
+  private static final String COUNT_AGGREGATE = "COUNT(*) AS \"total_records\"";
```
ferenc-csaky: Let's move this to the top.

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/server/jdbc/OffsetPageInfoQuery.java
```diff
@@ -0,0 +1,197 @@
...
+      PageFields fields,
+      boolean needsNextPage,
+      boolean needsEventTimes,
+      boolean needsTotals) {
+    this.page = page;
+    this.fields = fields;
+    this.needsNextPage = needsNextPage;
+    this.needsEventTimes = needsEventTimes;
+    this.needsTotals = needsTotals;
+  }
```
ferenc-csaky: Let's add a lombok ctor annotation instead: `@RequiredArgsConstructor(access = AccessLevel.PRIVATE)`

ferenc-csaky (changes requested):
Mostly looks good, added 2 minor comments

ferenc-csaky (approved):

PR #2195, opened 2026-07-08
feat: Update Flink to 2.3.0

ferenc-csaky:
## Key Changes
- Bumped Flink to 2.3.0
- For now set `table.exec.sink.require-on-conflict` to `false` to keep pre 2.3 behavior (we should adapt to this on a follow-up change)
- Adapted our code to the changes in `org.apache.flink.sql.parser.ddl.*` classes
- Renamed `FlinkSqlNodeFactory` to `FlinkSqlNodes` and refactored it a bit
- Updated SQL table property management to make sure we sort them for more deterministic output
- Adapted snapshots
- Removed some dead code


sqrl-planner/src/main/java/com/datasqrl/planner/FlinkPhysicalPlan.java
```diff
@@ -183,20 +182,19 @@ private void add(SqlNode node, String nodeSql, int batchIdx) {
...
-              formats.add(sqlTableOption.getValueString());
-          }
-        }
+        table
+            .getProperties()
+            .forEach(
+                (key, val) -> {
+                  switch (key) {
+                    case FlinkConnectorConfigWrapper.CONNECTOR_KEY:
+                      connectors.add(val);
```
mateczagany, line 191: Missing `break` here

mateczagany (changes requested):

mateczagany (approved):

PR #2213, opened 2026-07-14
fix: Add `no_source` hint to prevent a sink-only table marked as source

ferenc-csaky:
## Key Chnages
- Introduce `no_source` table hint to prevent marking a sink-only table as a source, which can lead to Flink planning errors
- Improve error message in case of unbounded source failure in batch mode
- Add a DAG planner and usecase test that covers the added logic


mateczagany (approved):

ferenc-csaky:
Superseded by #2218

PR #2218, opened 2026-07-16
feat: Support referencing individual tables from SQRL scripts in `IMPORT` and `EXPORT`

ferenc-csaky:
## Key Changes
- Improve `parseScript` with filterable output, so we can easily collect `CREATE TABLE` statements from scripts
- When an individual table is referenced from a `ScriptSqrlModule`, load the tables in a lazy manner (once), and return it if it's defined in the script
- Created `FlinkCalciteParser` and moved the Flink calcite parser init there, cause now we need that in two places
- Added proper error message when an internal table is referenced from another script
- Added DAG planner tests to cover the added logic
- Updates some usecases to cover both `IMPORT` and `EXPORT` usage

## TODO
- [x] Add test(s) for `IMPORT`
- [x] Update docs


sqrl-planner/src/main/java/com/datasqrl/engine/stream/flink/FlinkCalciteParser.java
```diff
@@ -0,0 +1,62 @@
...
+  private static CalciteParser getCalciteParser(Optional<TableEnvironmentImpl> tEnv) {
+    try {
+      var tEnvImpl = tEnv.orElse(TableEnvironmentImpl.create(new Configuration()));
+      var calciteSupplierField = ParserImpl.class.getDeclaredField("calciteParserSupplier");
+      calciteSupplierField.setAccessible(true);
+      return ((Supplier<CalciteParser>) calciteSupplierField.get(tEnvImpl.getParser())).get();
+    } catch (NoSuchFieldException | IllegalAccessException e) {
+      throw new RuntimeException(e);
+    }
+  }
```
mateczagany, line 55: This should also work:

```suggestion
  private static CalciteParser getCalciteParser(Optional<TableEnvironmentImpl> tEnv) {
    var tEnvImpl = tEnv.orElseGet(() -> TableEnvironmentImpl.create(new Configuration()));
    return ((PlannerBase) tEnvImpl.getPlanner()).createFlinkPlanner().parser();
  }
```

mateczagany (approved):

PR #2230, opened 2026-07-23
feat: Introduce lazy JBang JAR build

ferenc-csaky:
## Motivation

If there are JBang UDFs in a project, rebuilding the JBang JAR in every `compile` is a huge waste of time, cause in most cases, those logic do not change at all. This feature helps with keeping the built JAR around and will only rebuild when any source JBang file gets changed. So if a project is actively developed compile time won't be punished by the long JBang JAR build.

## Key Changes
- `JBangPreprocessor` now checks whether `jbang-udfs.jar` is present in the directory where it finds JBang source Java files, and skips rebuilding when the existing JAR contains matching SHA-256 hashes for all passed JBang files. The existing JAR will then be picked up by the `JarPreprocessor`.
- When `JBangRunner` builds a JAR, it now cleans up that JAR and performs extra steps:
  - Adds one `MANIFEST.MF` named entry per JBang source file with a `SHA-256-Digest` attribute containing that file's SHA-256 hash.
  - Adds the proper service loader descriptors for the JBang UDFs under `META-INF/services`, so when the JAR is picked up by the `JarPreprocessor` it can register them properly.
- Once `jbang-udfs.jar` is built, it is also copied back to the same source directory where the JBang scripts are placed, so it is kept between multiple `compile` calls.

### DIFF
--- a/sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
+++ b/sqrl-cli/src/main/java/com/datasqrl/packager/preprocess/JBangPreprocessor.java
@@ -15,15 +15,19 @@
  */
 package com.datasqrl.packager.preprocess;
 
-import com.datasqrl.config.PackageJson;
+import com.datasqrl.error.ErrorCollector;
 import com.datasqrl.packager.FilePreprocessingPipeline;
 import com.datasqrl.util.FilenameAnalyzer;
 import com.datasqrl.util.JBangRunner;
+import com.google.common.hash.Hashing;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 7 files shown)


mbroecheler:
@ferenc-csaky One idea: rather than doing an expiration, you could check if last modified of the jar is bigger or equal to the last modified of the java file. I think that would be more robust and lead to less "I updated the java but it's not working" user errors.

velo:
or, record a txt file with checksums for java files

ferenc-csaky:
@mbroecheler if by "last modified" you mean the filesystem last modified date, that's not robust as well. With mounting into to docker imgs on different OS-es, there is no guarantee that's intact.

@velo Checksum can work, but instead of a separate txt file, I'd put that into `MANIFEST.MF`. Polluting the project source with some generated junk can be annoying and confusing too.

mateczagany (approved):

PR #2235, opened 2026-07-27
refactor: Rationalize SQL dialect and statement code, to simplify adding new dialects

ferenc-csaky:
## Key Changes

- Centralized dialect SQL translation discovery and dispatch.
- Made translation maps immutable and locale-safe.
- Consolidated relational-plan and SQL-string conversion into `SqlConverters`.
- Replaced duplicated dialect converter pairs with shared conversion logic.
- Added service-based lookup for dialect SQL converters.
- Updated JDBC statement factories to use the consolidated converter.

## Snapshot Diffs
1. I turned back on identifier quoting in Snowflake
2. Some postgres ocnverters were instantiated multiple times before, that causes the generated index diff


mbroecheler (approved):
Looks good. I did not validate the snapshot changes.

mateczagany (approved):

PR #2236, opened 2026-07-27
feat: Derive pg_partman partition width from ttl unit and partition-divisor setting

WellMafra:
Reworked per review feedback from @ferenc-csaky and @mbroecheler: instead of adding a PG-specific second argument to the engine-agnostic `ttl` hint, the pg_partman partition width is now derived automatically from the TTL duration and a postgres-specific `partition-divisor` engine setting.

## How the partition width is picked

For range-partitioned tables with `ttl(X unit)` and the postgres setting `partition-divisor` (`P`, default 100):

```
X_min     = TTL normalized to minutes
Floor_min = 1 unit (as declared in the ttl hint) in minutes
menu      = [15 min, 30 min, 1h, 2h, 4h, 6h, 8h, 12h, 1d, 2d, 4d, 1w, 2w, 4w, 8w, 12w]  # calendar-aligned
W_target  = max(X_min / P, Floor_min)
W         = largest menu value <= W_target, else smallest menu value
```

The divisor caps the partition count for long TTLs, while the unit the user declares the TTL with sets the floor: `ttl(14 days)` produces 1-day partitions, `ttl(48 hours)` produces hourly partitions. Retention is still always the TTL duration.

## Key Changes
- `TtlHint` stays a single-argument hint and now captures the declared unit. Units are validated to be between minute and week (`min`, `hour`, `day`, `week` + short/plural forms); seconds and below or months and above raise `StatementParserException` with the source location. The `cache(...)` hint is unchanged.
- New postgres engine setting `partition-divisor` (default `100`), read in `PostgresJdbcEngine` and validated as a positive integer.
- `AbstractJdbcStatementFactory` gains an overridable `derivePartitionInterval(partitionType, ttl, ttlUnit)` hook (null by default); `PostgresStatementFactory` implements the formula above for RANGE tables.
- `CreateTableJdbcStatement.partitionInterval` now holds the computed width (still omitted from JSON when null); `PgPartmanExtension` uses it directly and the old TTL-bucketing `deriveInterval` fallback is removed.
- Documented the unit constraint in `sqrl-language.md` and the `partition-divisor` setting + partitioning behavior in `configuration-engine/postgres.md`.
- Tests: `TtlHintTest` rewritten for unit parsing/validation, new `PostgresStatementFactoryTest` covering the formula (floor dominance, divisor scaling, menu snapping, smallest-entry fallback, config parsing), updated `PgPartmanExtensionTest`, and `partitionTest.sqrl` extended with a `ttl(48 hours)` table (snapshot regenerated).

## Behavior change
Derived partition widths change for existing TTLs: e.g. `ttl(30 days)` previously produced `1 week` partitions and now produces `1 day` (floor-driven), per the agreed default handling.


ferenc-csaky (commented):
I don't think this solution is viable as is. The `ttl` hint is orthogonal to the external system, and now we are introducing a PG-specific parameter to it, which violates the contract of this hint, i.e. it works regardless what external data system connection that table has. Cause TTL is also meaningful for Kafka for example.

It makes sense to control this on the table level, but i may add a new `pg_partman` hint for this maybe?

Or another idea to extend the `engine` hint to engine-specific configs, which is a more general construct, but would solve this problem too.

cc @mbroecheler

mbroecheler:
@WellMafra I agree with @ferenc-csaky reasoning. We don't want this engine specific logic in the generic TTL hint. Moreover, I feel like it is too much to ask our users to fine-tune this variable. Instead, we should have better default handling that works for 99% of cases. Here is what I propose:

We have a postgres specific `partition-divisor` setting. Let's call that number `P`. When a user configured `ttl(X unit)` -meaning a TTL of X with the given time unit - then we set the partition width `W` with the same time unit using the following formula:
```
X_min = normalize(X, unit)
Floor_min = normalize(1, unit)
menu = [15 min, 30min, 1h, 2h, 4h, 6h, 8h, 12h, 1d, 2d, 4d, 1w, 2w, 4w, 8w, 12w]        # ensure calendar-alignment
W_target = max(X_min / P, Floor_min)
W = largest menu value (in min) <= W_target, else smallest menu value
```

This allows the user to control the number of partitions (via config option) and the floor (via the unit they use).
We should validate that the unit used for the TTL is at least `min` and at most `week`.

The default value for `partition-divisor` should be 100.

That means, a `ttl(14 days)` would result in a partition width of 1 day (driven by the Floor_min).

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/AbstractJdbcStatementFactory.java
```diff
@@ -197,6 +196,15 @@ protected PartitionType getPartitionType(
     return partitionKey.isEmpty() ? PartitionType.NONE : PartitionType.LIST;
   }
 
+  /**
+   * The width of the partitions for range-partitioned tables with a TTL - null when the dialect
+   * does not support time-based partitioning.
+   */
+  protected String derivePartitionInterval(
```
ferenc-csaky: I'd change this to `Optional<String>` instead, and add an `.orElse(null)` at L175. That's a more explicit and clear contract.

sqrl-planner/src/main/java/com/datasqrl/planner/hint/TtlHint.java
ferenc-csaky, line 1: The biggest time unit Flink's `TimeUtils` can handle are `DAYS` now, so it will fail to parse anything above that. This means handling weeks can be omitted.

Also, at this point I'd rather create our own `com.datasqrl.util.TimeUtils` in `sqrl-planner` that is built on top of Flink's `TimeUtils`, so it should be able to parse a duration, and also extract the defined unit as a `ChronoUnit`. So we don't pollute this class with general duration parsing code.

sqrl-planner/src/test/java/com/datasqrl/engine/database/relational/AbstractJdbcStatementFactoryTest.java
ferenc-csaky, line 1: Delete this, with the DAGPlanner tests in place, this is redundant.

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresStatementFactory.java
```diff
@@ -46,8 +49,47 @@
 
 public class PostgresStatementFactory extends AbstractJdbcStatementFactory {
 
+  public static final String PARTITION_DIVISOR_KEY = "partition-divisor";
+  public static final int DEFAULT_PARTITION_DIVISOR = 100;
```
ferenc-csaky: We should not hardcode this default value, but add this setting to `default-package.json`, so it will always present. We will also need to add this field to `packageSchema.json` as an integer type.

Then we can simplify `parsePartitionDivisor` to:
```java
return Integer.parseInt(engineConfig.getSetting("partition-divisor"));
```

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresStatementFactory.java
```diff
@@ -92,6 +134,32 @@ protected PartitionType getPartitionType(
...
+    var floorMinutes = ttlUnit.getDuration().toMinutes();
+    var targetMinutes = Math.max(ttl.toMinutes() / (double) partitionDivisor, floorMinutes);
+    var interval = PARTITION_MENU_LABELS[0];
+    for (var i = 0; i < PARTITION_MENU_MINUTES.length; i++) {
+      if (PARTITION_MENU_MINUTES[i] <= targetMinutes) {
+        interval = PARTITION_MENU_LABELS[i];
+      }
+    }
+    return interval;
+  }
```
ferenc-csaky: I'd eliminate this into it's own class. This logic is pretty specific and would be better to put it into a separate class `PostgresPartitionInterval`, and we can add a `from` or `of` factory method that calculates this.

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresStatementFactory.java
```diff
@@ -46,8 +49,47 @@
...
+   * Calendar-aligned partition widths pg_partman can use, in minutes, with their interval labels.
+   */
+  private static final long[] PARTITION_MENU_MINUTES = {
+    15, 30, 60, 120, 240, 360, 480, 720, 1440, 2880, 5760, 10080, 20160, 40320, 80640, 120960
+  };
+
+  private static final String[] PARTITION_MENU_LABELS = {
+    "15 minutes", "30 minutes", "1 hour", "2 hours", "4 hours", "6 hours", "8 hours", "12 hours",
+    "1 day", "2 days", "4 days", "1 week", "2 weeks", "4 weeks", "8 weeks", "12 weeks"
+  };
```
ferenc-csaky: Instead of 2 arrays, it would be simpler to define this as a `Map<Integer, String>`, and just use the Guava map builder to init.

ferenc-csaky (changes requested):

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/AbstractJdbcStatementFactory.java
WellMafra: done

sqrl-planner/src/main/java/com/datasqrl/planner/hint/TtlHint.java
WellMafra: done

sqrl-planner/src/test/java/com/datasqrl/engine/database/relational/AbstractJdbcStatementFactoryTest.java
WellMafra: done

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresStatementFactory.java
WellMafra: done

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresStatementFactory.java
WellMafra: done

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresStatementFactory.java
WellMafra: done

mbroecheler:
@WellMafra Can we please rename the postgres engine setting for this to `partition-ttl-divisor` to make it more explicit as to what it applies to? Please make sure this is added to the documentation as well. Thank you!

sqrl-planner/src/main/java/com/datasqrl/util/TimeUtils.java
```diff
@@ -0,0 +1,58 @@
...
+package com.datasqrl.util;
+
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
+
+/** Duration parsing utilities built on top of Flink's {@link org.apache.flink.util.TimeUtils}. */
+public final class TimeUtils {
+
+  private TimeUtils() {}
```
ferenc-csaky: Let's use Lombok's `@NoArgsConstructor(access = AccessLevel.PRIVATE)` annotation to spare the ctor.

sqrl-planner/src/main/java/com/datasqrl/planner/hint/TtlHint.java
```diff
@@ -65,16 +107,20 @@ public static Duration parseDuration(ParsedObject<SqrlHint> source) {
           "%s hint only supports one duration argument (e.g. `2 days`).",
           source.get().name());
     }
+    return parseDurationArgument(source, arguments.get(0));
+  }
+
+  private static Duration parseDurationArgument(ParsedObject<SqrlHint> source, String argument) {
```
ferenc-csaky: I do not really see why we split here, `parseDurationArgument` is only called by `parseDuration`, so IMO it was reasonable that it was originally defined as 1 method. Maybe the back and forth going on in this class resulted in this.

But `parseDuration` is now only used by the `CacheHint`, so lets move it there in its original form.

WellMafra:
@ferenc-csaky @mbroecheler 
Changes applied.

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresPartitionInterval.java
ferenc-csaky, line 1: Keeping `interval` as the state seems to be weird like how we use this now, cause we just need the string right away. So I'd suggest to make this a stateless util class that simply returns the `String`. I also removed the necessity of getting the first element with the iterator which is an okay thing, but we can spare it:
```java
public static String asString(
    Duration ttl, ChronoUnit ttlUnit, int partitionTtlDivisor) {

  var floorMinutes = ttlUnit.getDuration().toMinutes();
  var targetMinutes = Math.max(ttl.toMinutes() / (double) partitionTtlDivisor, floorMinutes);

  String interval = null;
  for (var entry : PARTITION_MENU.entrySet()) {
    if (interval == null || entry.getKey() <= targetMinutes) {
      interval = entry.getValue();
    }
  }

  return interval;
}
```

ferenc-csaky (changes requested):
Mostly looks good, added some final comments. Thanks!

sqrl-planner/src/main/java/com/datasqrl/util/TimeUtils.java
WellMafra: done

sqrl-planner/src/main/java/com/datasqrl/planner/hint/TtlHint.java
WellMafra: done

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/PostgresPartitionInterval.java
WellMafra: done

ferenc-csaky (approved):

PR #2238, opened 2026-07-29
refactor: Extract deployment file model classes to `sqrl-deployment-model`

ferenc-csaky:
## Summary

Extract deployment-file models into a dependency-light `sqrl-deployment-model` module. Deployment JSON is now produced from explicit file models rather than planner physical plans, keeping compiler state and persisted contracts separate.

## Motivation

Deployment-file consumers should not require Calcite, Flink, or planner implementation dependencies. Persisting physical plans also exposes compiler-only state and can accidentally serialize unsupported runtime types.

## Implementation

- Add the `sqrl-deployment-model` Maven module, limited to JDK types and Jackson annotations.
- Add `CombinedPlanModel`, `FlinkPlanModel`, `JdbcPlanModel`, `JdbcStatementModel`, `KafkaPlanModel`, `KafkaNewTopicModel`, and `MutationDatabaseModel`.
- Add `EnginePhysicalPlan.toFileModel()` and update `Packager` to serialize that representation.
- Add explicit mappings for Flink, JDBC, Kafka, and combined database plans.
- Preserve the combined-plan `plans` wrapper while mapping every nested plan to its file model. This prevents Jackson from traversing Calcite `RelDataType` instances during deployment serialization.
- Keep planner lifecycle behavior in `sqrl-planner`; `KafkaNewTopic` remains the `EngineCreateTable` wrapper around the deployment DTO.
- Keep mutation-database creation and compatibility checks in planner helpers while persisting `MutationDatabaseModel`.
- Update config loaders and CLI execution to consume deployment models when reading `postgres.json`, `kafka.json`, and mutation-database output.

## Compatibility

- Preserve existing deployment JSON structure, including the combined `plans` wrapper.
- Preserve JDBC CREATE TABLE metadata: fields, primary key, partition key/type/count, and TTL.
- Preserve Kafka topic convenience constructors and defaults.
- Mark derived `KafkaPlanModel.isEmpty()` as ignored so `kafka.json` only contains `topics` and `testRunnerTopics`.
- Retain Jackson deserialization support for persisted `TestPlan.jdbcViews` as `GenericJdbcStatement`.

## Testing

- Expand `PlanModelSerializationTest` to cover JDBC generic, CREATE TABLE, and extension statements; Flink fields; Kafka topic mapping/defaults; Kafka JSON properties; and combined-plan serialization containing a Calcite `BIGINT` type.
- Add a CLI regression test for deserializing persisted `TestPlan.jdbcViews`.


sqrl-deployment-model/src/main/java/com/datasqrl/deployment/model/JdbcStatementModel.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  public enum PartitionType {
+    NONE,
+    HASH,
+    LIST,
+    RANGE
+  }
+
+  public record Field(String name, String type, boolean nullable, String description) {}
```
velo, line 58: This is component-for-component identical to `JdbcStatement.Field`, and `JdbcPhysicalPlan.toStatementModel` maps between them field by field.

Since `Field` is already pure data with no planner dependencies, moving the existing record into this module and having `JdbcStatement` reference it would delete both the duplicate type and the mapping loop. Same question applies to `MutationDatabaseModel.Table`/`TableDefinition`/`ColumnDefinition`, which are straight copies of the originals.

The cost of keeping both is that every future field is two edits plus a mapper line, and forgetting the mapper line compiles fine and silently drops the field from the deployment file.

velo (commented):
I had an overlapping PR (#2237) doing the same extraction with a different technique, and I'm closing mine in favour of this one. Two things yours does better that mine didn't: flattening `GenericJdbcStatement` + `CreateTableJdbcStatement` into a single `JdbcStatementModel` (which incidentally fixes a real round-trip bug — `JdbcStatement` carries `@JsonDeserialize(as = GenericJdbcStatement.class)`, so reading `postgres.json` today silently drops `ttl`/`primaryKey`/`partitionKey`/`partitionType`/`numPartitions` on every CREATE TABLE), and keeping the published model free of compiler state.

The comments below are the parts where I think my approach held up better — mainly `@JsonValue`-based delegation instead of an explicit `toFileModel()` hook. Take or leave them; the module split itself is right either way.

velo (approved):

PR #2241, opened 2026-07-29
feat: Add Spark SQL as a shallow query engine

ferenc-csaky:
## Key Changes
- Add the `sparksql` query engine with Spark SQL statement generation backed by Iceberg deployment artifacts.
- Add Spark SQL Calcite conversion and dialect support, including backtick identifiers and Spark-native scalar, temporal, array, map, and struct type rendering.
- Refactor JDBC DDL generation to quote table, view, index, and column identifiers through the active SQL dialect.
- Make query-engine server configuration optional so engines without runtime server configuration, such as Spark SQL, are supported.
- Add a Spark SQL compile use case and snapshot covering primitive, temporal, binary, collection, row, and aggregate result types.

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/calcite/dialect/ExtendedSparkSqlDialect.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/calcite/dialect/ExtendedSparkSqlDialect.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright © 2021 DataSQRL (contact@datasqrl.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datasqrl.calcite.dialect;
+
+import org.apache.calcite.config.NullCollation;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.sql.SqlAlienSystemTypeNameSpec;
+import org.apache.calcite.sql.SqlDataTypeSpec;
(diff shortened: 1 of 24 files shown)


sqrl-testing/sqrl-testing-integration/src/test/resources/engine-validation/package-sparksql-no-server.json
```diff
@@ -0,0 +1,14 @@
+{
+  "version": "1",
+  "enabled-engines": ["flink", "iceberg", "sparksql"],
+  "script": {
+    "main": "script.sqrl"
+  },
+  "engines": {
+    "snowflake": {
```
mateczagany, line 8: `snowflake` config is not needed here

sqrl-planner/src/main/java/com/datasqrl/calcite/dialect/ExtendedSparkSqlDialect.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.calcite.sql.SqlDialect;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.dialect.SparkSqlDialect;
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+public class ExtendedSparkSqlDialect extends SparkSqlDialect {
+
+  public static final Context DEFAULT_CONTEXT =
+      SqlDialect.EMPTY_CONTEXT
+          .withDatabaseProduct(DatabaseProduct.SPARK)
```
mateczagany, line 31: `SqlDialect.EMPTY_CONTEXT` will default to high null collation, while Spark's would be low.

sqrl-testing/sqrl-testing-integration/src/test/resources/engine-validation/package-sparksql-no-server.json
ferenc-csaky: this gets removed in the Redshift PR anyways, so i'll leave it as is.

sqrl-planner/src/main/java/com/datasqrl/calcite/dialect/ExtendedSparkSqlDialect.java
ferenc-csaky: Good catch, ty!

mateczagany (approved):

PR #2246, opened 2026-07-31
feat: Support defining sort order in `index` hints

ferenc-csaky:
## Summary

Adds sort-order support for JDBC index hints and propagates Calcite sort direction into generated B-tree index candidates.

## Key Changes

- Supports index hints such as `/*+ index(BTREE, col_a DESC, col_b) */`.
- Adds index-column directions to `IndexDefinition`.
- Adds sort metadata to `QueryIndexSummary`.
- Preserves Calcite `RelFieldCollation.Direction` when selecting indexes for limited sorted queries.
- Generates distinct index names for descending columns.
- Emits `DESC` in PostgreSQL `CREATE INDEX` DDL.
- Adds `IndexType.supportsSortOrder()` to centralize index capability checks.
- Rejects descending columns for index types that do not support sort order.
- Adds parser, index-selector, and PostgreSQL DDL tests.

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/plan/global/QueryIndexSummary.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/plan/global/QueryIndexSummary.java
@@ -53,18 +53,13 @@ public class QueryIndexSummary {
   private static final QueryIndexSummary EMPTY =
       new QueryIndexSummary(null, Set.of(), Set.of(), Set.of(), 1.0);
 
-  public static final String INDEX_NAME = "_index_";
-
   @Include NamedTable table;
   @Include Set<Integer> equalityColumns;
   @Include Set<Integer> inequalityColumns;
   @Include Set<IndexableFunctionCall> functionCalls;
 
-  // TODO: add support for sort orders
-  // List<IndexableSort> sorts;
-
   /** Keeps track of the relative frequency of query conjunctions as we reduce them */
-  double count = 1.0;
+  double count;
 
   public static List<QueryIndexSummary> ofFilter(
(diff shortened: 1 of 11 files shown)


sqrl-planner/src/main/java/com/datasqrl/plan/global/QueryIndexSummary.java
```diff
@@ -53,18 +53,13 @@ public class QueryIndexSummary {
...
       new QueryIndexSummary(null, Set.of(), Set.of(), Set.of(), 1.0);
 
-  public static final String INDEX_NAME = "_index_";
-
   @Include NamedTable table;
   @Include Set<Integer> equalityColumns;
   @Include Set<Integer> inequalityColumns;
   @Include Set<IndexableFunctionCall> functionCalls;
 
-  // TODO: add support for sort orders
```
mateczagany, line 63: The current code does not preserve the sort direction. In [IndexSelector line 438](https://github.com/DataSQRL/sqrl/blob/804db9ed041a141d0a052d14d6b785d1d6593159/sqrl-planner/src/main/java/com/datasqrl/plan/global/IndexSelector.java#L438) the direction is not used from the collation

ferenc-csaky: This is intentional. We only allow the `DESC` sort order when it is explicitly defined in an index hint. `QueryIndexSummary` and `IndexSelector` does not apply it automatically, cause in most cases DB engines can optimize for that better.

mateczagany (approved):

PR #2265, opened 2026-08-06
feat: Cost-based DuckDB Iceberg scan reuse

ferenc-csaky:
## Summary

Add ability to adaptively materialize repeated DuckDB Iceberg table scans as `MATERIALIZED` CTEs.

## Key Changes

- Groups repeated scans by source table within one query.
- Uses table cardinality and scan count to decide whether materialization is worthwhile.
- Factors shared deterministic direct filter predicates into the CTE source.
- Keeps original branch filters unchanged.
- Adds `scan-cte-cardinality-divisor` config under `duckdb` engine to tune the threshold. (Defaults to 1)


mbroecheler (approved):
Very nice. Would be great to add a usecase test to test this with cardinality annotations on the tables and some crazy generated query that hits the tables multiple times. As an end2end test.

sqrl-planner/src/main/java/com/datasqrl/engine/database/relational/DuckDbMaterializedScanCtePlanner.java
```diff
@@ -0,0 +1,172 @@
...
+    return ctes;
+  }
+
+  boolean shouldMaterialize(Deque<TableScan> scans) {
+    if (scans.size() < 2) {
+      return false;
+    }
+
+    var scan = scans.getFirst();
+    var cardinality = scan.getCluster().getMetadataQuery().getRowCount(scan);
```
mateczagany: Should this not come from `tableIdMap.get(tableId).tableAnalysis().getTableStatistic()`? I think this one will always just return the placeholder value of 1e8.

ferenc-csaky: Yes it should -.- I explicitly refactored this once, and seems like in one iteration that dummy LLM reverted it to its original impl. I just recognized this too during the test case add...good eye ;)

PR #2267, opened 2026-08-06
docs: Document postgres data-checksums deployment setting

WellMafra:
## What was done
- Added `data-checksums` to the `engines.postgres.deployment` example in the cloud deployment configuration docs
- Added a "Data Checksums" section documenting the setting: defaults to `true` (the PostgreSQL 18 default), applied at initdb time, immutable after the database is created

## Why it was done
- Documents the new `engines.postgres.deployment.data-checksums` knob introduced in DataSQRL/cloud-compilation#1094


documentation/docs/configuration-engine/cloud-deployment.md
```diff
@@ -278,6 +279,28 @@ Set it to `false` to bootstrap the database **tables-only**, skipping all index
 }
 ```
 
+## Data Checksums (`data-checksums`)
+
+Controls whether PostgreSQL data-page checksums are enabled for the database. Defaults to `true` (the PostgreSQL 18 default). PostgreSQL only.
```
velo, line 284: considering this configuration goes into `engines/postgres` make sense that it is postgreSQL only, no need to mention.

Would be nice a link to this feature on postgres side, for people wanting to know more.

velo (approved):

PR #2268, opened 2026-08-07
fix!: Enforce a single `ROWTIME` for `SqrlDefinition`s

ferenc-csaky:
## Summary

Prevents SQRL table definitions from silently selecting a ROWTIME column when a query propagates multiple Flink event-time attributes.

- Rejects result schemas with multiple ROWTIME columns and reports the table and conflicting columns.
- Excludes processing-time attributes from ROWTIME detection.
- Skips the validation only when the final result is a `FOR SYSTEM_TIME AS OF` temporal join.
- Continues validation if a temporal join is wrapped by an ordinary join.
- Adds the `MULTIPLE_ROWTIME_COLUMNS` error code, remediation guidance, and a regression fixture.

## Resolution

Keep one authoritative ROWTIME column unchanged. Materialize every other ROWTIME column as a regular timestamp by casting it to its own type:

```sql
CAST(o.eventTime AS TIMESTAMP_LTZ(3)) AS createdAt,
l.eventTime AS updatedAt
```
For joins, suggest to derive the authoritative timestamp from the latest input event time when appropriate:
```sql
GREATEST(o.eventTime, l.eventTime) AS eventTime
```


sqrl-planner/src/main/java/com/datasqrl/util/CalciteUtil.java
```diff
@@ -77,6 +77,10 @@ public static boolean isNestedTable(RelDataType type) {
   }
 
   public static boolean isRowTime(RelDataType type) {
```
mbroecheler, line 79: should we call this `isEventTime` to be most clear?

sqrl-planner/src/main/resources/com/datasqrl/error/errorCodes/multiple_rowtime_columns.md
```diff
@@ -0,0 +1,9 @@
+DataSQRL can propagate only one Flink event-time attribute as the table's ROWTIME column.
```
mbroecheler: This table has multiple rowtime columns which makes he authoritative timestamp for the records in this table ambiguous. This can cause confusion for downstream users and unexpected data processing results.

Keep the authoritative rowtime column as-is and `CAST` all other rowtime columns to a regular timestamps:
*current Example*
For joins, you usually take the `GREATEST` of all rowtimes from the joined tables:
*Join example*

sqrl-testing/sqrl-testing-integration/src/test/resources/dagplanner/comprehensiveTest.sqrl
```diff
@@ -18,8 +18,8 @@ ExplicitDistinct := SELECT customerid, `timestamp`, name FROM (SELECT *, (ROW_NU
 --maps to explicit distinct (check rowtype)
 InvalidDistinct := SELECT customerid, `timestamp`, name AS namee FROM (SELECT *, (ROW_NUMBER() OVER (PARTITION BY customerid ORDER BY `timestamp` DESC)) AS _rownum FROM Customer) WHERE (_rownum = 1);
 
-MissedTemporalJoin := SELECT * FROM ExternalOrders o JOIN ExplicitDistinct c ON o.customerid = c.customerid;
-TemporalJoin := SELECT * FROM ExternalOrders o JOIN ExplicitDistinct FOR SYSTEM_TIME AS OF `time` c ON o.customerid = c.customerid;
+MissedTemporalJoin := SELECT o.*, CAST(`timestamp` AS TIMESTAMP_LTZ(3)) AS `timestamp`, name FROM ExternalOrders o JOIN ExplicitDistinct c ON o.customerid = c.customerid;
+TemporalJoin := SELECT o.*, CAST(`timestamp` AS TIMESTAMP_LTZ(3))  AS `timestamp`, name FROM ExternalOrders o JOIN ExplicitDistinct FOR SYSTEM_TIME AS OF `time` c ON o.customerid = c.customerid;
```
mbroecheler: on a temporal join it should mark anything from the 'c' table as rowtime, does it?

sqrl-testing/sqrl-testing-integration/src/test/resources/graphql-validation/comprehensiveTest.sqrl
```diff
@@ -15,8 +15,8 @@ ExplicitDistinct := SELECT customerid, `timestamp`, name FROM (SELECT *, (ROW_NU
 --maps to explicit distinct (check rowtype)
 InvalidDistinct := SELECT customerid, `timestamp`, name AS namee FROM (SELECT *, (ROW_NUMBER() OVER (PARTITION BY customerid ORDER BY `timestamp` DESC)) AS _rownum FROM Customer) WHERE (_rownum = 1);
 
-MissedTemporalJoin := SELECT * FROM ExternalOrders o JOIN ExplicitDistinct c ON o.customerid = c.customerid;
-TemporalJoin := SELECT * FROM ExternalOrders o JOIN ExplicitDistinct FOR SYSTEM_TIME AS OF `time` c ON o.customerid = c.customerid;
+MissedTemporalJoin := SELECT o.*, CAST(`timestamp` AS TIMESTAMP_LTZ(3)) AS `timestamp`, name FROM ExternalOrders o JOIN ExplicitDistinct c ON o.customerid = c.customerid;
+TemporalJoin := SELECT o.*, CAST(`timestamp` AS TIMESTAMP_LTZ(3))  AS `timestamp`, name FROM ExternalOrders o JOIN ExplicitDistinct FOR SYSTEM_TIME AS OF `time` c ON o.customerid = c.customerid;
```
mbroecheler: same here

mbroecheler (changes requested):
Looks good. Minor improvement to message and question about one test.

sqrl-testing/sqrl-testing-integration/src/test/resources/dagplanner/comprehensiveTest.sqrl
ferenc-csaky: For any even-time temporal join, the rowtime attribute must come from the left side table.

sqrl-planner/src/main/java/com/datasqrl/util/CalciteUtil.java
ferenc-csaky: technically `ROWTIME` == event time,  for wall-clock time, it uses `PROCTIME`. that's why i gave a more general name to the original method.

sqrl-testing/sqrl-testing-integration/src/test/resources/dagplanner/comprehensiveTest.sqrl
mbroecheler: Exactly, and `timestamp` comes from the right side (ExplicitDistinct), so why do you need to cast it?

ferenc-csaky: Cause the validation does not care about it's context, and the schema still contains 2 timestamp fields marked as `ROWTIME`. But I added a detect logic which skips the validation if the final SELECT is coming from a temporal join

mateczagany (approved):

PR #2269, opened 2026-08-07
feat: Support project-wide Kafka `SOURCE_WATERMARK` config

### LINKED ISSUE #2193: Integrate `SOURCE_WATERMARK` to the Kafka engine
Now SQRL can handle watermark intervals for Kafka tables based on `package.json` configs.

Now that we have a Kafka `SOURCE_WATERMARK` implementation on the runner side, we should integrate that into SQRL as well.

That would make it possible to define source watermark strategy project-wide, and then it won't be necessary to add it to each table by hand.

ferenc-csaky:
## Summary
Adds project-wide Kafka configuration to emit Flink `SOURCE_WATERMARK()` clauses for Kafka timestamp metadata columns.

## Key Changes
- Adds `use-source-watermark` config option for regular Kafka sources.
- Adds `use-transaction-source-watermark` config option for transactional Kafka sources.
- Defaults both settings to false, preserving bounded-delay watermark behavior.
- Generates `WATERMARK FOR <timestamp-column> AS SOURCE_WATERMARK()` when enabled.
- Declares the new settings in the package schema and rejects unknown Kafka engine settings.
- Adds SQL-node coverage and updates the complex-mutation compilation snapshot for source watermarks.


mateczagany (commented):
`SOURCE_WATERMARK()` only works on sources implementing `SupportsSourceWatermark`, which `flink-connector-kafka` doesn't. So with the default `'connector' = 'kafka'` this compiles and submits fine, then dies ~2s in with `TableException: SOURCE_WATERMARK() is a declarative marker function and doesn't have concrete runtime implementation`. 

Can we make it so it fails at compilation instead, not after job submission?

ferenc-csaky:
Our `(upsert)-kafka-safe` variants support it, but yes, we should validate this cause as of now it's possible to use the normal Kafka connectors as well.

PR #2276, opened 2026-08-12
feat: Make GraphQL parser limits configurable via vertx-config.json

velo:
## Problem

Large GraphQL queries are rejected by the server with:

```
More than 15,000 'grammar' tokens have been presented.
To prevent Denial Of Service attacks, parsing has been cancelled.
```

That ceiling is graphql-java's `ParserOptions.MAX_QUERY_TOKENS` default. It was hard-coded — nothing in `vertx-config.json` could change it, and because `ServerConfig` is `@JsonIgnoreProperties(ignoreUnknown = true)`, adding a key by hand would deserialize silently and do nothing.

## Change

Adds a `graphQLParserConfig` section to `vertx-config.json` exposing all four graphql-java parser limits:

```json
"graphQLParserConfig" : {
  "maxCharacters" : 1048576,
  "maxTokens" : 15000,
  "maxWhitespaceTokens" : 200000,
  "maxRuleDepth" : 500
}
```

Defaults are read from graphql-java's own constants, so behaviour is unchanged when the section is absent or left at its defaults.

The limits are applied in `HttpServerVerticle.bootstrap()` rather than `GraphQLServerVerticle`, because they are JVM-wide statics and that verticle is deployed once per model version. Both `setDefaultParserOptions` and `setDefaultOperationParserOptions` are set — query parsing resolves its options from the *operation* defaults, so setting only the generic ones would not have fixed the error. They are applied with `transform(...)` on the existing defaults so graphql-java's other operation-specific settings (`captureIgnoredChars`, `redactTokenParserErrorMessages`, …) are preserved. SDL parsing keeps its own, much higher, defaults and is deliberately untouched.

Overrides flow through the existing config plumbing, so they can be set from `package.json` under `engines.vertx.config.graphQLParserConfig`.

Note: no Vert.x-native options class carries these limits (`GraphQLHandlerOptions` only has `requestBatchingEnabled` / `requestMultipartEnabled`), so this follows the existing repo pattern for DataSQRL-owned config sections, matching `GraphQLTailSampleTracingConfig`.

## Testing

- `GraphQLParserConfigTest` — defaults match graphql-java, both parser defaults updated, SDL defaults untouched, other parser settings preserved, no-op when left at defaults.
- `ServerConfigTest` — the new section deserializes, defaults on absent/null.
- `ServerFunctionContainerIT` — end-to-end: with `maxTokens: 5` in `vertx-config.json` the server rejects a query with `More than 5 'grammar' tokens have been presented`; the pre-existing test in the same class covers the default config still parsing that query fine.

Full `mvn test` is green; `ServerFunctionContainerIT` passes locally.


sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/server/config/GraphQLParserConfig.java
```diff
@@ -0,0 +1,84 @@
...
+  private int maxWhitespaceTokens = ParserOptions.MAX_WHITESPACE_TOKENS;
+  private int maxRuleDepth = ParserOptions.MAX_RULE_DEPTH;
+
+  /**
+   * Applies these limits to the JVM-wide graphql-java parser defaults. Both the generic and the
+   * operation defaults are updated because query parsing resolves its options from {@link
+   * ParserOptions#getDefaultOperationParserOptions()}. SDL parsing keeps its own, much higher,
+   * defaults and is deliberately left untouched.
+   */
+  public void applyToParserDefaults() {
```
ferenc-csaky: Let's call this `apply` or `applyParserOptions`, this name sounds weird.

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/server/config/GraphQLParserConfig.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  private boolean isDefault() {
+    return maxCharacters == ParserOptions.MAX_QUERY_CHARACTERS
+        && maxTokens == ParserOptions.MAX_QUERY_TOKENS
+        && maxWhitespaceTokens == ParserOptions.MAX_WHITESPACE_TOKENS
+        && maxRuleDepth == ParserOptions.MAX_RULE_DEPTH;
+  }
+
+  private ParserOptions withLimits(ParserOptions options) {
```
ferenc-csaky: nit: `withCustomConfig` describes more what this does

sqrl-server/sqrl-server-vertx-base/src/main/java/com/datasqrl/server/HttpServerVerticle.java
```diff
@@ -130,6 +130,10 @@ public void stop(Promise<Void> stopPromise) throws Exception {
   // ---------------------------------------------------------------------------
 
   private void bootstrap(Promise<Void> startPromise) {
+    // graphql-java parser limits are JVM-wide statics, so they are applied once here rather than
+    // per GraphQLServerVerticle, which is deployed once per model version
```
ferenc-csaky: Shrink or remove this comment, the details are available on the method javadoc

ferenc-csaky (changes requested):

ferenc-csaky (approved):

PR #2279, opened 2026-08-12
feat: Execute standalone extension statements in local runner

WellMafra:
## What was done
- `DatasqrlRun.initPostgres` now executes the plan's `standaloneExtensionStatements` (e.g. the pg_partman `CREATE EXTENSION` + `create_parent` block) after the regular statements, non-fatally: a failure logs a warning and the run continues, so a Postgres without the extension behaves exactly as before.
- Install `postgresql-${POSTGRES_VERSION}-partman` in the sqrl-cli base image so `run`/`test` can actually create the extension locally.

## Why it was done
- Tables with a `ttl()` hint are emitted as `PARTITION BY RANGE` parents with no inline default partition (pg_partman owns their lifecycle), but the local runner never executed the partman block — leaving the parent with zero partitions, so every insert fails with `no partition of relation ... found for row`. This made `sqrl test`/`sqrl run` unusable for any project using `ttl()` on a Postgres table.
- Validated end-to-end against DataSQRL/cloud-backend metrics: with this image the runner logs `Executing standalone extension statement partman of type EXTENSION`, historical test fixtures land in the partman DEFAULT partition, and all 14 snapshot tests pass.

### DIFF
--- a/sqrl-cli/Dockerfile.base
+++ b/sqrl-cli/Dockerfile.base
@@ -22,6 +22,7 @@ RUN apt-get update \
             postgresql-${POSTGRES_VERSION} \
             postgresql-contrib-${POSTGRES_VERSION} \
             postgresql-${POSTGRES_VERSION}-pgvector \
+            postgresql-${POSTGRES_VERSION}-partman \
     && curl -1sLf 'https://dl.redpanda.com/nzc4ZYQK3WRGd9sy/redpanda/cfg/setup/bash.deb.sh' | bash \
     && apt-get install -y --no-install-recommends redpanda \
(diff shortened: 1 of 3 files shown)


ferenc-csaky:
Why executing `standaloneExtensionStatements` AFTER the regular statements? I would consider an extension setup more general than a `CREATE TABLE`, so in an initialization phase, I'd rather execute `standaloneExtensionStatements` first.

WellMafra:
> Why executing `standaloneExtensionStatements` AFTER the regular statements?

because the create_parent requires the parent table to already exist.

ferenc-csaky (approved):

PR #2281, opened 2026-08-14
feat: Generate and serve OpenAPI deployment artifacts

ferenc-csaky:
## Key Changes

- Generate a versioned OpenAPI specification during compilation and package it as a deployment artifact.
- Support an optional OpenAPI specification in API configuration and validate generated specifications for backwards compatibility.
- Serve the compiled OpenAPI artifact at runtime and update its server URL from the incoming request host.
- Fall back to generating OpenAPI with the request host when the artifact is unavailable or invalid.
- Derive and persist GraphQL operation response schemas for OpenAPI response documentation.
- Refactor server-model generation into `ServerModelManager`.
- Split URI parameter extraction into query and path handling.
- Make POST operations handle path parameters and JSON body parameters.
- Reject POST requests when a path parameter conflicts with the corresponding body value.

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/server/ServerModelManager.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/server/ServerModelManager.java
@@ -16,25 +16,33 @@
 package com.datasqrl.server;
 
 import com.datasqrl.config.PackageJson;
+import com.datasqrl.config.WorkspacePaths;
 import com.datasqrl.engine.server.ServerPhysicalPlan;
 import com.datasqrl.error.ErrorCollector;
+import com.datasqrl.server.config.OpenApiConfig;
+import com.datasqrl.server.config.ServletConfig;
 import com.datasqrl.server.converter.GraphQLSchemaConverter;
 import com.datasqrl.server.converter.GraphQLSchemaConverterConfig;
 import com.datasqrl.server.graphql.RootGraphQLModel;
 import com.datasqrl.server.graphql.RootGraphQLModel.StringSchema;
+import com.datasqrl.server.openapi.OpenApiService;
 import com.datasqrl.server.operation.ApiOperation;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import lombok.AllArgsConstructor;
+import org.openapitools.openapidiff.core.OpenApiCompare;
+import org.openapitools.openapidiff.core.model.ChangedOpenApi;
 import org.springframework.stereotype.Component;
 
 /** Generates the model for the server */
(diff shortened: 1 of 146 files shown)


sqrl-server/sqrl-server-core/src/main/java/com/datasqrl/server/operation/ApiOperation.java
```diff
@@ -40,6 +39,7 @@ public class ApiOperation {
   String uriTemplate;
 
   ResultFormat format;
+  @JsonIgnore ResultDefinition result;
```
mateczagany: The `@JsonIgnore` is no-op

sqrl-planner/src/main/java/com/datasqrl/server/ServerModelManager.java
```diff
@@ -91,4 +97,50 @@ public RootGraphQLModel generateGraphQLModel(ApiSources api, ServerPhysicalPlan
...
+      throw errors.exception(
+          "Failed to compare generated OpenAPI specification with %s: %s",
+          configuredOpenApi, e.getMessage());
+    }
+
+    errors.checkFatal(
+        !diff.isIncompatible(),
+        "The generated OpenAPI specification is not backwards compatible with %s: %s",
+        configuredOpenApi,
+        diff);
```
mateczagany: `ChangedOpenApi#toString` will be pretty ugly to read. Could probably use [ConsoleRender](https://github.com/OpenAPITools/openapi-diff/blob/63850d8074452f5761ce7e7e460307f0f127647a/core/src/main/java/org/openapitools/openapidiff/core/output/ConsoleRender.java#L22) of openapi-diff for better results.

sqrl-planner/src/main/java/com/datasqrl/server/ServerModelManager.java
```diff
@@ -91,4 +97,50 @@ public RootGraphQLModel generateGraphQLModel(ApiSources api, ServerPhysicalPlan
         .schema(schema)
         .build();
   }
+
+  public String generateOpenApiJson(ApiSources api, RootGraphQLModel model) {
+    var openApiService =
+        new OpenApiService(
```
mateczagany, line 107: We should use the merged server config here (the one writtent ot `vertx-config.json`) instead of the default `OpenApiConfig` and `ServletConfig` like in `RestBridgeVerticle` when constructing `OpenApiService`.

Currently, a project that customizes `engines.vertx.config` might get an artifact whose paths and title don't match the actual registered routes

PR #2288, opened 2026-08-19
feat: Configurable pg_partman premake and historical start partition

WellMafra:
## What was done
- Added `partition-premake` config key under `engines.postgres` (default 4) controlling `p_premake` in the generated pg_partman DDL
- Emit `p_start_partition => (now() - interval '<retention>')::text` in `create_parent` so historical partitions covering the retention window are created at setup time
- Passed the postgres `EngineConfig` through `DatabaseTableExtension.getDdl` so extensions can read their own settings
- Updated JSON schema, default package, docs, unit tests, and the DAGPlanner snapshot

## Why it was done
- `p_premake => 4` was hardcoded, always pre-creating 4 future daily partitions with no way to tune it
- Without `p_start_partition` no historical partitions exist, so catch-up replay after a blue-green deploy lands the full retention window in the DEFAULT partition, requiring a lock-heavy post-catch-up sweep and forcing full DEFAULT scans on every new partition attach

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/function/translation/postgres/extensions/PgPartmanExtension.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/function/translation/postgres/extensions/PgPartmanExtension.java
@@ -15,6 +15,8 @@
  */
 package com.datasqrl.function.translation.postgres.extensions;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
 import com.datasqrl.config.PackageJson.EngineConfig;
 import com.datasqrl.deployment.model.JdbcStatementModel.PartitionType;
 import com.datasqrl.engine.database.relational.CreateTableJdbcStatement;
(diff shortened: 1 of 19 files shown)


sqrl-planner/src/main/java/com/datasqrl/function/translation/postgres/extensions/PgPartmanExtension.java
```diff
@@ -23,18 +24,29 @@
...
 
   @Override
-  public String getDdl(Collection<CreateTableJdbcStatement> createTables) {
+  public String getDdl(
+      Collection<CreateTableJdbcStatement> createTables, EngineConfig engineConfig) {
+    var premake =
+        Integer.parseInt(
+            engineConfig.getSetting(PARTITION_PREMAKE_KEY, Optional.of(DEFAULT_PREMAKE)));
+    Preconditions.checkArgument(
+        premake >= 1, "%s must be a positive number", PARTITION_PREMAKE_KEY);
```
ferenc-csaky: This class should not know about `EngineConfig` (and its content, so the added constants should be removed as well. Move the validation to `PostgresStatementFactory`, and pass the premake value only to `getDdl`.

And the default will come through `default-package.json` so we don't need to hardcode it in any Java class.

ferenc-csaky (approved):

PR #2289, opened 2026-08-19
feat: Add Flink 2.3 insert conflict handling

ferenc-csaky:
## Summary

- Generate Flink 2.3 `ON CONFLICT` clauses only for upsert-capable sinks.
- Preserve append-only sinks, including `print`, by omitting conflict clauses.
- Carry optional `InsertConflictStrategy.ConflictBehavior` through relational and table analysis.
- Mark `TO_CHANGELOG` with explicit `DEDUPLICATE` behavior without changing its `TableType`.
- Centralize conflict behavior resolution in `FlinkConflictBehaviorUtil`.
- Use explicit behavior when present, otherwise resolve from table type and source watermarks.
- Move `SqlScriptPlannerUtil` to the planner utility package.

## Conflict Resolution Policy

- Append-only sinks: no `ON CONFLICT` clause.
- Explicit analyzed behavior: use Flink's `ERROR`, `NOTHING`, or `DEDUPLICATE` behavior.
- `STATE`, `VERSIONED_STATE`, and `STATIC` tables: `ON CONFLICT DO DEDUPLICATE`.
- `STREAM` tables with watermarked leaf sources: `ON CONFLICT DO NOTHING`.
- Other `STREAM` tables: `ON CONFLICT DO DEDUPLICATE`.
- Other table types without explicit behavior: no clause.

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/planner/util/FlinkConflictBehaviorUtil.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/planner/util/FlinkConflictBehaviorUtil.java
@@ -18,10 +18,13 @@ package com.datasqrl.planner.util;
 import com.datasqrl.planner.analyzer.RelNodeAnalysis;
 import com.datasqrl.planner.analyzer.TableAnalysis;
 import com.datasqrl.planner.analyzer.TableOrFunctionAnalysis;
+import com.datasqrl.planner.tables.SourceSinkTableAnalysis;
+import com.datasqrl.planner.tables.SqrlTableFunction;
 import java.util.List;
 import java.util.Optional;
 import org.apache.flink.sql.parser.dml.SqlInsertConflictBehavior;
 import org.apache.flink.table.api.InsertConflictStrategy;
+import org.apache.flink.table.catalog.ResolvedSchema;
 
 /**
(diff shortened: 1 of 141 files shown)


sqrl-planner/src/main/java/com/datasqrl/planner/util/FlinkConflictBehaviorUtil.java
```diff
@@ -0,0 +1,121 @@
...
+  private static boolean hasSourceWatermarks(TableOrFunctionAnalysis table) {
+    if (table instanceof TableAnalysis tableAnalysis
+        && !tableAnalysis.isSourceOrSink()
+        && !tableAnalysis.getFromTables().isEmpty()) {
+
+      return tableAnalysis.getFromTables().stream()
+          .allMatch(FlinkConflictBehaviorUtil::hasSourceWatermarks);
+    }
+
+    return table.getRowTime().isPresent();
```
mateczagany: This will always return empty, as `relNode` will always be null here: https://github.com/DataSQRL/sqrl/blob/aec5d1d29336e2cad8b6c54b1686f6b08493a3af/sqrl-planner/src/main/java/com/datasqrl/planner/analyzer/TableAnalysis.java#L187

As a result, none of the test snapshots actually contain `ON CONFLICT DO NOTHING`.

mateczagany (approved):

PR #2293, opened 2026-08-20
fix: Analyze process table function inputs

### LINKED ISSUE #2056: Process Table Functions not analyzed correctly
When a table definition contains a process table function (PTF) in the FROM clause, it does not get correctly analyzed and the tables that the PTF references are not extracted as sources. This causes subtrees in the DAG to be missing.

Solution: List all tables referenced by the PTF as sources in the logical rel node analysis.

ReguiguiMohamed:
Fixes #2056

Process table function scans could skip analysis of their relational inputs, leaving referenced source tables out of the logical plan analysis and causing missing DAG subtrees.

Analyze each input and rebuild the table function scan with the analyzed nodes so its source dependencies are preserved.
So my Addition is simple : `getInputAnalyses(functionScan)` now visits every relational input, allowing existing table-scan analysis to collect its sources. The scan is then rebuilt with the analyzed inputs so those dependencies remain in the logical plan and DAG.

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/planner/analyzer/SQRLLogicalPlanAnalyzer.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/planner/analyzer/SQRLLogicalPlanAnalyzer.java
@@ -447,7 +447,13 @@ public class SQRLLogicalPlanAnalyzer implements SqrlRelShuttle {
       }
     }
     // Generic table function call
-    return setProcessResult(RelNodeAnalysis.builder().relNode(functionScan).build());
+    var inputs = getInputAnalyses(functionScan);
+    return setProcessResult(
+        RelNodeAnalysis.builder()
+            .relNode(
+                updateRelnode(
+                    functionScan, inputs.stream().map(RelNodeAnalysis::getRelNode).toList()))
+            .build());
   }
 
   @Override
--- a/sqrl-planner/src/main/java/com/datasqrl/planner/analyzer/SQRLLogicalPlanAnalyzer.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/planner/analyzer/SQRLLogicalPlanAnalyzer.java
@@ -92,7 +92,9 @@ import org.apache.calcite.rex.RexSubQuery;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.validate.SqlUserDefinedTableFunction;
 import org.apache.calcite.tools.RelBuilder;
+import org.apache.flink.table.functions.FunctionKind;
 import org.apache.flink.table.planner.calcite.FlinkRelBuilder;
+import org.apache.flink.table.planner.functions.bridging.BridgingSqlFunction;
 import org.apache.flink.table.planner.functions.sql.SqlWindowTableFunction;
 import org.apache.flink.table.planner.plan.schema.TableSourceTable;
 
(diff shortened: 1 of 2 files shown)


sqrl-planner/src/main/java/com/datasqrl/planner/analyzer/SQRLLogicalPlanAnalyzer.java
```diff
@@ -447,7 +447,13 @@ public RelNode visit(TableFunctionScan functionScan) {
...
     }
     // Generic table function call
-    return setProcessResult(RelNodeAnalysis.builder().relNode(functionScan).build());
+    var inputs = getInputAnalyses(functionScan);
+    return setProcessResult(
+        RelNodeAnalysis.builder()
+            .relNode(
+                updateRelnode(
+                    functionScan, inputs.stream().map(RelNodeAnalysis::getRelNode).toList()))
+            .build());
```
ferenc-csaky: Instead of changing the generic path, I'd rather restrict this behavior for PTFs, cause it's not necessary to traverse scan inputs for simple table functions:
```java
    } else if (call.getOperator() instanceof BridgingSqlFunction function
        && function.getDefinition().getKind() == FunctionKind.PROCESS_TABLE) {
      // PTFs carry TABLE arguments as relational inputs that form DAG dependencies.
      var inputs = getInputAnalyses(functionScan);
      return setProcessResult(
          RelNodeAnalysis.builder()
              .relNode(
                  updateRelnode(
                      functionScan, inputs.stream().map(RelNodeAnalysis::getRelNode).toList()))
              .build());
    }
    // Generic table function call
    return setProcessResult(RelNodeAnalysis.builder().relNode(functionScan).build());
```

ferenc-csaky (changes requested):
Thanks for the contribution!

See my added comment, and applying this change will also require to update the snapshot of the below use-case:
```
sqrl-testing/sqrl-testing-integration/src/test/resources/usecases/to-changelog
``` 
That can be done by removing the snapshot file `sqrl-testing/sqrl-testing-integration/src/test/resources/snapshots/com/datasqrl/UseCaseCompileTest/to-changelog-package.txt` and then running:
```
sqrl-testing/sqrl-testing-integration/src/test/java/com/datasqrl/UseCaseCompileTest.java
```

sqrl-planner/src/main/java/com/datasqrl/planner/analyzer/SQRLLogicalPlanAnalyzer.java
ReguiguiMohamed: hey @ferenc-csaky thanks for the review, glad to be contributing here

so i narrowed it to PTFs as you suggested, the generic path is back to what it was
on main. Regenerated to-changelog-package.txt: `AppendStream` now lists `InputAgg`
under Inputs. It also collapsed `AppendStream `into the `InputAgg `GraphQL type,
I'm here if theres anything else to be corrected in this PR !

ferenc-csaky (approved):
LGTM! I'll let the CI run, if it fails that's because of it does not handle forks properly, but the change and the snapshot looks good, so I'll merge it.

ReguiguiMohamed:
> LGTM! I'll let the CI run, if it fails that's because of it does not handle forks properly, but the change and the snapshot looks good, so I'll merge it.

Truly an honour! I'll make sure to continue using and improving this app! It really saved me a lot of time in my work

PR #2316, opened 2026-08-27
docs: document the postgres cpu-limit deployment setting

velo:
Documentation companion to DataSQRL/cloud-compilation#1270, which adds `engines.postgres.deployment.cpu-limit`.

## What the setting does

Raises — or removes — the CPU ceiling of the PostgreSQL pods without changing the CPU request. By default the ceiling is the "Max CPU Burst" multiple of the request, which for every size except `dev` means limit == request: the database is throttled at exactly the CPU it reserves, even when the node has idle cores.

Three forms are documented: `"3x"` (a multiple of the request, so it moves with `instance-size`), `"6000m"` / `"6"` (absolute), and `"unlimited"` (no limit at all).

The new "CPU Limit" section also states the two consequences a user needs before setting it: pod QoS drops from Guaranteed to Burstable, and changing the value restarts the database because CNPG rolls the cluster on a pod-spec change.

## Also in this PR

The **Memory (GiB)** column of the PostgreSQL instance-size table was wrong for `dev`, `small` and `medium` — it listed double the memory those sizes actually get. Corrected to match `PostgresSize` in cloud-compilation (`dev` 2, `small` 4, `medium` 8); `large` and `xlarge` were already right. Verified against a running `small` database, whose postgres container has a 4Gi memory request and limit.

### DIFF
--- a/documentation/docs/configuration-engine/cloud-deployment.md
+++ b/documentation/docs/configuration-engine/cloud-deployment.md
@@ -138,6 +138,7 @@ PostgreSQL deployments consist of one primary instance and a configurable number
         "replica-count": 1,             // Number of read replicas (0 or larger)
         "disk-size-gb": 256,            // Disk size in GB (1 or larger)
         "auto-expand-percentage": 0.2,  // Auto-expand threshold (0 to disable, must be < 1)
+        "cpu-limit": "3x",              // CPU ceiling above the request (see "CPU Limit" below)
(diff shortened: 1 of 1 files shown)


velo:
Closing per discussion with Matthias: `cpu-limit` stays an advanced setting that cloud-compilation supports but that we don't document for end users yet, until we've experimented with it. The implementation lands in DataSQRL/cloud-compilation#1270.

PR #2319, opened 2026-08-27
fix: Resolve Flink insert conflicts after changelog inference

ferenc-csaky:
## Problem
Generated Flink inserts applied `ON CONFLICT` behavior before the planner determined whether each query produced inserts only or a changelog. This could add unnecessary clauses for append-only outputs, causing Flink planning errors, or omit required conflict handling when final sink planning found an upsert key incompatible with the sink primary key.

## Changes
- Add `FlinkInsertConflictPlanner` to resolve generated `ON CONFLICT` clauses from optimized query changelog modes.
- Preserve explicit conflict behavior and omit clauses for confirmed insert-only queries.
- Run final Flink validation for streaming plans with generated inserts, including runtime deployment plans that do not retain a compiled plan.
- Retry final Flink compilation with automatic conflict behavior when sink upsert-key validation requires it.
- Preserve the rendered statement set in unresolved Flink compilation errors.
- Encapsulate conflict recovery in the insert planner and update affected compilation snapshots.


sqrl-planner/src/main/java/com/datasqrl/planner/FlinkInsertConflictPlanner.java
```diff
@@ -0,0 +1,298 @@
...
+      default -> Optional.empty();
+    };
+  }
+
+  /**
+   * Returns whether the exception chain contains Flink's upsert-key validation failure for the
+   * given sink.
+   */
+  private static boolean hasUpsertKeyConflict(Throwable error, ObjectIdentifier targetTableId) {
+    var qualifiedTarget = "'" + targetTableId + "'";
```
mateczagany: `ObjectIdentifier#toString` will add backticks to this, while the Flink planner will not (`FlinkChangelogModeInferenceProgram`). It uses `ObjectIdentifier.asSummaryString()`, which does not have backticks.

I think you can check for this instead of the `containsAny` call below:

```
Strings.CS.contains(msg, "'" + targetTableId.asSummaryString() + "'")
```

PR #2338, opened 2026-09-02
feat: Validate enabled engine names

### LINKED ISSUE #2337: package.json schema accepts unknown engine names under `engines`, silently ignoring misplaced config
## Summary

`packageSchema.json` closes every engine object (`additionalProperties: false`) and closes the schema root, but the `engines` node itself declares only `type`, `minProperties` and `properties` — no `additionalProperties`, which defaults to `true` in JSON Schema.

The effect is asymmetric:

- an unknown key **inside** a known engine is rejected with a clear, fatal error
- an unknown **engine name** is accepted, ignored, and copied verbatim into `build/package.json`

## Impact

The common form of this mistake is a `deployment` block written one level too high:

```json
{
  "enabled-engines": ["flink"],
  "engines": {
    "flink": { "config": { "execution.runtime-mode": "BATCH" } },
    "deployment": { "schedule": { "cron": "*/10 * * * *", "timezone": "UTC" } }
  }
}
```

This validates today. It is parsed as configuration for an engine named `deployment`, which is not in `enabled-engines`, so it is never instantiated — engines are created from `enabled-engines` plus defaults (`ExecutionEnginesHolder#getEngines`) and each engine pulls its own subtree by name (`EnginesConfigImpl#getEngineConfig`). Nothing enumerates the keys of `engines` to check them.

The compile succeeds, the block survives the merge into `build/package.json`, and any downstream consumer of `engines.flink.deployment` sees nothing configured. For DataSQRL Cloud that means deployment settings — sizing, dedicated nodes, schedule — silently do not apply, with no warning anywhere in the compile output. A misplaced schedule in particular produces a batch pipeline that deploys and runs correctly once, then never runs again.

## Reproduction

Validating the document above against `packageSchema.json` with draft 2020-12 (the dialect `ConfigLoaderUtils#isValidJson` uses):

```
as-is, current schema                      : VALID (no errors)
+ "additionalProperties": false on engines : ['engines']: Additional properties are not allowed ('deployment' was unexpected)
control: "deploymentt" inside flink        : ['engines','flink']: Additional properties are not allowed ('deploymentt' was unexpected)
```

The control line shows the mistake is already caught one level deeper — only the top level of `engines` is open.

## Suggested fix

Add `"additionalProperties": false` to the `engines` node in `sqrl-planner/src/main/resources/jsonSchema/packageSchema.json`.

Before doing so, reconcile the allowed engine list with the engine factories. The schema currently lists `flink`, `duckdb`, `redshift`, `sparksql`, `snowflake`, `vertx`, `postgres`, `kafka`, while the registry also defines:

- `iceberg` — `IcebergEngineFactory.ENGINE_NAME`
- `postgres_log` — `PostgresLogEngineFactory.ENGINE_NAME`

Neither is configured by any `package.json` in this repo, so closing `engines` breaks no test, but it would start rejecting valid configurations that set them. Worth adding a test that a stray top-level key under `engines` fails validation, so the two lists cannot drift apart again.

## Priority

Low. Correct configurations are unaffected and the workaround is to place the block correctly. The value is turning a silent misconfiguration into a compile error.

ferenc-csaky:
## Summary
- Restrict `enabled-engines` to a non-empty array of supported engine names.
- Reject unknown engine names and scalar engine configuration values in schema tests.
- Reject scalar values for all `SqrlConfig.asList` calls instead of silently treating them as empty lists.
- Add schema and configuration regression coverage for invalid engine names and non-array list values.
- Update the config-loader fixture to use valid iceberg engine configuration.

## Testing
```
mvn test -pl sqrl-planner -am -Dtest=PackageJsonSchemaTest,ConfigLoaderUtilsTest,SqrlConfigTest -Dsurefire.failIfNoSpecifiedTests=false
```

Fixes #2337

### DIFF
--- a/sqrl-planner/src/main/resources/jsonSchema/packageSchema.json
+++ b/sqrl-planner/src/main/resources/jsonSchema/packageSchema.json
@@ -15,16 +15,23 @@
     },
     "enabled-engines": {
       "description": "List of engines to enable for the data pipeline. The compiler builds deployment artifacts for the enabled engines during the build.",
-      "oneOf": [
-        {
-          "type": "array",
-          "items": { "type": "string" },
-          "minItems": 1
-        },
-        {
-          "type": "string"
-        }
-      ]
+      "type": "array",
+      "items": {
+        "type": "string",
+        "enum": [
+          "duckdb",
+          "flink",
+          "iceberg",
+          "kafka",
+          "postgres",
+          "print",
+          "redshift",
+          "snowflake",
+          "sparksql",
+          "vertx"
+        ]
+      },
+      "minItems": 1
     },
     "compiler": {
       "type": "object",
@@ -352,7 +359,8 @@
           },
           "additionalProperties": false
         }
-      }
+      },
+      "additionalProperties": false
     },
     "connectors": {
       "type": "object",
--- a/sqrl-planner/src/test/java/com/datasqrl/config/PackageJsonSchemaTest.java
+++ b/sqrl-planner/src/test/java/com/datasqrl/config/PackageJsonSchemaTest.java
@@ -60,6 +60,8 @@ class PackageJsonSchemaTest {
   @ValueSource(
       strings = {
         "emptyEnabledEngines.json",
+        "invalidEnabledEngine.json",
+        "invalidEnabledEnginesType.json",
         "invalidVersionFormat.json",
         "emptyEnginesFlinkConnectors.json",
         "emptyTestRunner.json",
--- a/sqrl-planner/src/test/java/com/datasqrl/util/ConfigLoaderUtilsTest.java
+++ b/sqrl-planner/src/test/java/com/datasqrl/util/ConfigLoaderUtilsTest.java
@@ -149,7 +149,7 @@ class ConfigLoaderUtilsTest {
     assertThat(underTest).isNotNull();
     assertThat(underTest.getVersion()).isEqualTo(1);
 
-    assertThat(underTest.getEnabledEngines()).contains("test");
+    assertThat(underTest.getEnabledEngines()).containsExactly("iceberg");
 
     assertThat(underTest.getTestConfig()).isNotNull();
     assertThat(underTest.getEngines().getEngineConfig("flink")).isPresent();
--- a/sqrl-planner/src/main/java/com/datasqrl/config/SqrlConfig.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/config/SqrlConfig.java
@@ -271,9 +271,11 @@ public class SqrlConfig {
   public <T> Value<List<T>> asList(String key, Class<T> clazz) {
     var n = node();
     List<T> list = List.of();
-    if (n != null && n.has(key) && n.get(key).isArray()) {
+    if (n != null && n.has(key)) {
+      var value = n.get(key);
+      errors.checkFatal(value.isArray(), "Expected an array for key [%s]", getFullKey(key));
       list = new ArrayList<>();
-      for (JsonNode element : n.get(key)) {
+      for (var element : value) {
         list.add(MAPPER.convertValue(element, clazz));
       }
     }
--- a/sqrl-planner/src/test/java/com/datasqrl/config/SqrlConfigTest.java
+++ b/sqrl-planner/src/test/java/com/datasqrl/config/SqrlConfigTest.java
@@ -137,6 +137,15 @@ public class SqrlConfigTest {
   }
 
   @Test
+  void givenNonArrayValue_whenGetList_thenThrowsException() {
+    config.setProperty("list-value", "not-a-list");
+
+    assertThatThrownBy(() -> config.asList("list-value", String.class))
+        .isInstanceOf(CollectedException.class)
+        .hasMessageContaining("Expected an array for key [list-value]");
+  }
+
+  @Test
   void givenNestedProperties_whenGetSubConfig_thenReturnsNestedNalues() {
     var parentConfig = config.getSubConfig("parent");
     var childConfig = parentConfig.getSubConfig("child");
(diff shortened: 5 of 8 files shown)


sqrl-planner/src/main/resources/jsonSchema/packageSchema.json
```diff
@@ -15,16 +15,23 @@
...
-        {
-          "type": "array",
-          "items": { "type": "string" },
-          "minItems": 1
-        },
-        {
-          "type": "string"
-        }
-      ]
+      "type": "array",
```
velo, line 18: Dropping the string branch is right. But the root cause is one level down.

`SqrlConfig#asList` returns `List.of()` silently when the value is not an array. So `"enabled-engines": "flink"` compiled to zero engines with no error.

That applies to every list key, not just this one. And the final `loadResolvedConfig` in `loadUnresolvedConfig` passes a `null` schema, so it is not covered there.

Making `asList` fatal on a present-but-non-array value would fix the whole class.

sqrl-planner/src/main/resources/jsonSchema/packageSchema.json
```diff
@@ -15,16 +15,23 @@
...
+      "items": {
+        "type": "string",
```
velo, line 20: `ServiceLoaderDiscovery.findFirst` matches with `equalsIgnoreCase`.

So `["Flink"]` works today. A JSON Schema `enum` is case-sensitive, so this becomes a hard error.

Probably fine, just flagging it as a behaviour change.

sqrl-planner/src/main/resources/jsonSchema/packageSchema.json
```diff
@@ -15,16 +15,23 @@
...
+          "flink",
+          "iceberg",
+          "kafka",
+          "postgres",
+          "print",
+          "redshift",
+          "snowflake",
+          "sparksql",
+          "vertx"
+        ]
```
velo, line 32: This makes the schema a second source of truth.

Engine names are owned by `EngineFactory#getEngineName` + `@AutoService`, resolved via `ServiceLoaderDiscovery`. Adding an engine now also means editing this file. Forgetting it gives an enum error for a working engine.

Two ways out:

1. Validate where the name is already resolved. `ExecutionEnginesHolder#getEngineInstance` already rejects unknown names, just with a raw `ServiceLoaderException`. An `errors.fatal` listing `ServiceLoaderDiscovery.getAll(EngineFactory.class)` names cannot drift.
2. Keep the enum for editor completion, and add a test asserting it equals the SPI-discovered set.

Worth noting there are two more copies of this list already: `TestEngine.java:157-166` (missing `sparksql`, `redshift`, `print`, has a phantom `test`) and `documentation/docs/configuration.md:19-25`.

sqrl-planner/src/main/resources/jsonSchema/packageSchema.json
```diff
@@ -15,16 +15,23 @@
...
+          "duckdb",
+          "flink",
+          "iceberg",
+          "kafka",
+          "postgres",
+          "print",
+          "redshift",
+          "snowflake",
+          "sparksql",
+          "vertx"
```
velo, line 31: `postgres_log` is missing.

`PostgresLogEngineFactory.ENGINE_NAME` is `@AutoService`-registered, so it resolves at runtime. The schema now rejects it.

Its `plan` and `createTable` throw `UnsupportedOperationException`, so excluding it may well be deliberate. Just want it to be a decision, not an oversight.

Glad to see `print` in here. That one would have broken `sqrl test`, since `ExecutionEnginesHolder#initEnabledEngines` writes it back into `build/package.json` under the TEST goal.

sqrl-planner/src/test/java/com/datasqrl/config/PackageJsonSchemaTest.java
```diff
@@ -60,6 +60,8 @@ void validConfigFile(String configFileName) {
   @ValueSource(
       strings = {
         "emptyEnabledEngines.json",
+        "invalidEnabledEngine.json",
+        "invalidEnabledEnginesType.json",
```
velo, line 64: `testForErrors` only asserts `errors.isFatal()`.

The value of this change is the message a user sees for a typo. Nothing here checks that the message names the offending value.

Worth asserting it.

sqrl-planner/src/test/java/com/datasqrl/util/ConfigLoaderUtilsTest.java
```diff
@@ -149,7 +149,7 @@ void givenSinglePath_whenLoadingUnresolvedConfig_thenOverridesDefaults() {
     assertThat(underTest).isNotNull();
     assertThat(underTest.getVersion()).isEqualTo(1);
 
-    assertThat(underTest.getEnabledEngines()).contains("test");
+    assertThat(underTest.getEnabledEngines()).contains("iceberg");
```
velo: The test is `...thenOverridesDefaults`.

With `"test"` (a non-default name) `contains` proved the override. With `"iceberg"` it would also pass if defaults were merged rather than replaced.

`containsExactly("iceberg")` restores the guarantee.

velo (commented):
Direction is right. The old `oneOf` string branch was a real trap.

A few inline notes. Nothing blocking except maybe the `postgres_log` one.

Separately: `enabled-engines` is now closed, but the sibling `engines` object still has no `additionalProperties: false`. That is #2337. Happy to take it in a follow-up.

sqrl-planner/src/main/resources/jsonSchema/packageSchema.json
ferenc-csaky: `postgres_log` is not usable at the moment, we keep most of the code around if we want to support it in a later point.

sqrl-planner/src/main/resources/jsonSchema/packageSchema.json
ferenc-csaky: we explicitly agreed that this list does not change too much so it's okay to hardcode these into the schema

PR #2348, opened 2026-09-05
perf: backport CALCITE-7053 and CALCITE-7422 garbage collection into a shadowed HepPlanner

velo:
## Summary

Flink 2.3 bundles Calcite 1.36.0, whose `HepPlanner` runs a full mark-and-sweep garbage collection (`BreadthFirstIterator.reachable` over the whole graph plus a sweep of the vertex set and the digest map) after every rule transformation, once more at the top of every graph iterator and periodically per program instruction. A profile of a large compile shows about a third of the whole Flink optimization pass spent in that bookkeeping, with no rule on the stack. Calcite fixed this upstream:

- **CALCITE-7053** (1.41): stop collecting garbage at the top of `getGraphIterator` (the 2006 FIXME) and only do it where the topological iterators need a garbage-free graph.
- **CALCITE-7422** (main, "large plan mode"): after a transformation, release only the vertices that became unreachable through the discarded vertex (`tryCleanVertices`, a recursive in-degree walk from the discarded vertex) instead of scanning the graph; drop the per-instruction amortised GC and the GC before the topological iterators, which are no-ops once the graph is kept garbage free; identity cache for shared sub-trees in `setRoot`; consistency assertions in trace mode.

This PR backports those two changes onto the 1.36.0 `HepPlanner` as a same-package shadowed copy in `sqrl-planner/src/main/java/org/apache/calcite/plan/hep/HepPlanner.java`. The parts of CALCITE-7422 that alter rule application order or common-subexpression merging (`HepVertexIterator` for ARBITRARY/DEPTH_FIRST, recursive parent merging in `contractVertices`, `firedRulesCache`) are deliberately left out so that plans stay byte-identical.

### How the shadowing works

`sqrl-planner` depends on `flink-table-planner_2.12` directly, so Calcite classes are on the application classpath (no planner loader). The repo already relies on the module's own classes preceding its dependencies, both in the surefire classpath and in the `sqrl-cli` shaded jar (first class wins; this is how `org.apache.flink.table.planner.loader.PlannerModule` is overridden today). The shadowed class keeps its ASF header, so `pom.xml` gains a license-check exclude for `org/apache/calcite/plan/hep/**` next to the existing `org/apache/flink/**` one. It imports Flink's relocated Guava (`org.apache.flink.calcite.shaded.com.google.common`) like the other shadowed Calcite classes in the module.

### Why plans are identical

Garbage collection never affects which rules fire on live vertices; what matters is that the topological iterators (`TOP_DOWN`/`BOTTOM_UP`, used by most Flink Hep programs) iterate `graph.vertexSet()` and therefore must never see garbage. The subtree GC keeps the invariant "no garbage after every transformation": the only edges a transformation removes are the discarded vertex's inward edges, so in a DAG the set of vertices that became unreachable has the discarded vertex as its only source and the in-degree walk from it collects exactly that set. A result that no parent adopted (converter-rule trait filtering) is added as a second candidate. The digest map is cleaned with `remove(digest, vertex)`, matching the sweep's value-based removal. Iteration order (DepthFirstIterator, TopologicalOrderIterator, reversed list for BOTTOM_UP) and `nTransformations` accounting are untouched.

Verified empirically: with a temporary check that ran the original full mark-and-sweep, `assertNoCycles` and `assertGraphConsistent` after every single transformation and asserted the graph size did not change, `UseCaseCompileTest`, `DAGPlannerTest` and `PredicatePushdownTest` pass (209 tests, 0 failures), and the benchmark artifacts (Flink SQL, compiled plan, schemas) hash identical to control.

## Benchmark

5 warm cycles of cloud-backend core/agent/metrics against the compilation server, one compile slot, same machine, artifacts hashed.

### Against main (control = unmodified main)

| module | variant | min | p50 | p80 | p95 | max | p50 vs control |
|---|---|---|---|---|---|---|---|
| core | control | 15.98 s | 16.37 s | 16.63 s | 17.26 s | 17.26 s | +0.0% |
| core | hep-planner | 15.32 s | 15.46 s | 15.62 s | 16.83 s | 16.83 s | -5.5% |
| agent | control | 0.83 s | 0.88 s | 0.92 s | 3.22 s | 3.22 s | +0.0% |
| agent | hep-planner | 0.85 s | 2.44 s | 2.84 s | 3.34 s | 3.34 s | +179.1% |
| metrics | control | 25.49 s | 26.92 s | 27.34 s | 28.01 s | 28.01 s | +0.0% |
| metrics | hep-planner | 20.63 s | 20.91 s | 21.44 s | 21.48 s | 21.48 s | -22.3% |

control: version=1.0-perf-control-SNAPSHOT artifacts identical to control
hep-planner: version=1.0-perf-hep-planner-SNAPSHOT artifacts identical to control

The agent module is bimodal in both variants (0.85 s vs 2.4-3.3 s per cycle) because it compiles right after core and pays that compile's heap; it spends almost no time in HepPlanner and the min values are equal, so it is noise, not an effect of this change.

### On top of the combined perf branch (`perf/combined-pr`)

| module | variant | min | p50 | p80 | p95 | max | p50 vs combined |
|---|---|---|---|---|---|---|---|
| core | combined | 4.91 s | 5.41 s | 5.51 s | 5.89 s | 5.89 s | +0.0% |
| core | hep-on-combined | 5.10 s | 5.28 s | 5.48 s | 8.48 s | 8.48 s | -2.4% |
| agent | combined | 0.48 s | 0.52 s | 0.54 s | 2.52 s | 2.52 s | +0.0% |
| agent | hep-on-combined | 0.47 s | 0.48 s | 0.57 s | 0.64 s | 0.64 s | -7.4% |
| metrics | combined | 7.62 s | 8.18 s | 8.32 s | 8.47 s | 8.47 s | +0.0% |
| metrics | hep-on-combined | 6.01 s | 6.73 s | 6.75 s | 8.01 s | 8.01 s | -17.6% |

combined: version=1.0-perf-combined-SNAPSHOT artifacts identical to combined
hep-on-combined: version=1.0-perf-hep-combined-SNAPSHOT artifacts identical to combined

## Test plan

- [x] `UseCaseCompileTest` 71/71, `DAGPlannerTest` 101/101, `PredicatePushdownTest` 37/37 (`mvn -pl sqrl-planner,sqrl-testing/sqrl-testing-integration test -Dtest=...`), snapshots unchanged
- [x] Same suites with the trace-mode GC cross-check forced on after every transformation: 0 assertion failures
- [x] Shaded `sqrl-cli.jar` contains the shadowed class (`javap` shows `tryCleanVertices`)
- [x] Benchmark artifacts identical to control

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1014,6 +1014,7 @@
               <header>com/mycila/maven/plugin/license/templates/APACHE-2.txt</header>
               <excludes>
                 <exclude>**/src/**/org/apache/flink/**</exclude>
+                <exclude>**/src/**/org/apache/calcite/plan/hep/**</exclude>
                 <exclude>**/src/test/resources/**</exclude>
                 <exclude>**/resources/templates/init-project/**</exclude>
                 <exclude>**/resources/templates/functions/**</exclude>
--- a/sqrl-planner/src/main/java/org/apache/calcite/plan/hep/HepPlanner.java
+++ b/sqrl-planner/src/main/java/org/apache/calcite/plan/hep/HepPlanner.java
@@ -1,0 +1,1106 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.calcite.plan.hep;
+
+import static java.util.Objects.requireNonNull;
+import static org.apache.calcite.linq4j.Nullness.castNonNull;
+import static org.apache.flink.calcite.shaded.com.google.common.base.Preconditions.checkArgument;
+
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Deque;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.IdentityHashMap;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
+import java.util.Set;
+import org.apache.calcite.linq4j.function.Function2;
+import org.apache.calcite.linq4j.function.Functions;
+import org.apache.calcite.plan.AbstractRelOptPlanner;
+import org.apache.calcite.plan.CommonRelSubExprRule;
+import org.apache.calcite.plan.Context;
+import org.apache.calcite.plan.RelDigest;
+import org.apache.calcite.plan.RelOptCost;
+import org.apache.calcite.plan.RelOptCostFactory;
+import org.apache.calcite.plan.RelOptCostImpl;
+import org.apache.calcite.plan.RelOptMaterialization;
+import org.apache.calcite.plan.RelOptPlanner;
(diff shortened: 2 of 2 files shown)


ferenc-csaky:
I'd say let's not do this until we must. Flink 2.4 already includes Calcite 1.41, which fixes one of the issues, and maybe it gets bumped to 1.42 before the release too.

PR #2364, opened 2026-09-10
fix: INSERT analysis should not mutate query

### LINKED ISSUE #2363: INSERT analysis mutates the SQL tree
The SQRL compiler emits duplicate ORDER BY clauses for sorted INSERT queries because validation mutates the SQL tree later reused for output. This breaks the compaction job when skipGlobalSort is false or omitted. The fix is to analyze a separate copy of the query.

mateczagany:
Fix duplicate ORDER BY clauses in generated SQL for sorted INSERT queries.

Calcite validation mutates the query tree in place, pushing ORDER BY and LIMIT/OFFSET into the inner SELECT. The compiler later emits the original INSERT, whose outer wrapper still contains those clauses, duplicating them. Analyze a separate copy to preserve the original query for output.

Extend the existing multi-batch-compile use case with sorted INSERTs, including LIMIT/OFFSET, and update its UseCase and DAG snapshots. Both snapshot tests fail without the fix and pass with it.

Fixes #2363

### DIFF
--- a/sqrl-testing/sqrl-testing-integration/src/test/resources/snapshots/com/datasqrl/DAGPlannerTest/insertInto.txt
+++ b/sqrl-testing/sqrl-testing-integration/src/test/resources/snapshots/com/datasqrl/DAGPlannerTest/insertInto.txt
@@ -64,9 +64,9 @@ LogicalAggregate(group=[{0}], numEntries=[COUNT()])
   LogicalProject(customerid=[$0])
     LogicalTableScan(table=[[default_catalog, default_database, Customer]])
 SQL:
-SELECT `Customer`.`customerid`, COUNT(*) AS `numEntries`
-FROM `default_catalog`.`default_database`.`Customer` AS `Customer`
-GROUP BY `Customer`.`customerid`
+SELECT `customerid`, COUNT(*) AS `numEntries`
+FROM `Customer`
+GROUP BY `customerid`
 >>>flink-sql-no-functions.sql
 CREATE TABLE `Customer` (
   `customerid` BIGINT NOT NULL,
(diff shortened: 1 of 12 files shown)


sqrl-testing/sqrl-testing-integration/src/test/resources/snapshots/com/datasqrl/DAGPlannerTest/insertInto.txt
```diff
@@ -114,9 +114,9 @@ SELECT *
 FROM `default_catalog`.`default_database`.`Customer`
 ;
 INSERT INTO `customer_id_count`
-SELECT `Customer`.`customerid`, COUNT(*) AS `numEntries`
-FROM `default_catalog`.`default_database`.`Customer` AS `Customer`
-GROUP BY `Customer`.`customerid`
+SELECT `customerid`, COUNT(*) AS `numEntries`
+FROM `Customer`
+GROUP BY `customerid`
```
ferenc-csaky, line 119: I think this holds a minor risk, cause now we don't use the full ID of the source table here, so I'd guess there can be edge-cases where this may try to read from a wrong DB (?).

ferenc-csaky (commented):
The logic itself looks good, I added 1 inline a comment about a risk. IMO it's worth merging like it is cause the copy logic is pretty clean and we do not try to defer the analysis mutation instead.

And to be completely safe against Flink catalog/DB switches, the `INSERT INTO` target should also be referenced with its full ID, which never happened previously as well.

So I'd be +1 to merge this and only change it to full ID reference when we stumble upon a bug about it. WDYT?

sqrl-testing/sqrl-testing-integration/src/test/resources/snapshots/com/datasqrl/DAGPlannerTest/insertInto.txt
mateczagany: I agree, this is a new risk with this PR. I will create an issue for that and try to come up with something.

ferenc-csaky (approved):

PR #2368, opened 2026-09-10
feat: Normalize engine configurations

ferenc-csaky:
## Key Changes

  - Remove configuration blocks for engines not present in the effective enabled-engines set.
  - Perform cleanup after `ExecutionEnginesHolder` expands test-only engines, preserving their default configurations.
  - Emit a warning only for removed engine configurations supplied by user package files.
  - Preserve silent cleanup for default-only engine configurations.
  - Add regression coverage for test-added engines and aggregated user configuration warnings.

### DIFF
--- a/sqrl-planner/src/main/java/com/datasqrl/util/ConfigLoaderUtils.java
+++ b/sqrl-planner/src/main/java/com/datasqrl/util/ConfigLoaderUtils.java
@@ -15,6 +15,7 @@
  */
 package com.datasqrl.util;
 
+import static com.datasqrl.config.PackageJsonImpl.ENGINES_PROPERTY;
 import static com.google.common.base.Preconditions.checkArgument;
 
 import com.datasqrl.config.PackageJson;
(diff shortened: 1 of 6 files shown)


sqrl-planner/src/main/java/com/datasqrl/util/ConfigLoaderUtils.java
```diff
@@ -301,32 +315,107 @@ static PackageJson loadUnresolvedConfig(
...
+      return;
+    }
+
+    var enabledEngineNames =
+        enabledEngines.valueStream().map(JsonNode::asText).collect(Collectors.toSet());
+
+    defaultEngineConfigs.stream()
+        .filter(engine -> !userEngineConfigs.contains(engine))
+        .filter(engine -> !enabledEngineNames.contains(engine))
+        .forEach(engine -> ((ObjectNode) engines).remove(engine));
```
mateczagany: In [ExecutionEnginesHolder:153](https://github.com/DataSQRL/sqrl/blob/f97329d18428c7d214dac2c32ddbde27379a2595/sqrl-planner/src/main/java/com/datasqrl/config/ExecutionEnginesHolder.java#L153) Postgres engine will be added automatically for `test` runs if none is specified, but this runs first, so the default configs will be removed anyways.

Not that bad, it's just these 2 configs, if engine `postgres` is not added in the package.json, and only affects `test`: 
```
      "partition-ttl-divisor": 100,
      "partition-premake": 4
```

mateczagany (approved):

PR #2370, opened 2026-09-11
docs: Document GraphQL and OpenAPI configuration

ferenc-csaky:
## Summary

  - Correct outdated GraphQL schema-generation guidance and document inferred schemas.
  - Clarify GraphQL, MCP, and REST operation generation and configuration.
  - Document generated OpenAPI artifacts, runtime OpenAPI/Swagger endpoints, and Vert.x OpenAPI settings.
  - Explain that `script.api.<version>.openapi` is a compatibility baseline rather than an API definition.
  - Add a configuration matrix covering inferred, single-version, versioned, and compatibility-checked APIs.

### DIFF
--- a/documentation/docs/interface.md
+++ b/documentation/docs/interface.md
@@ -9,7 +9,8 @@ Based on the SQRL script, DataSQRL generates the interface for the compiled data
 
 For data products, DataSQRL generates view definitions as deployment assets in `build/deploy/plan` which can be queried directly.
 
-The last three are APIs that can be invoked programmatically. The `protocols` [compiler configuration](configuration.md#compiler-compiler) controls which API protocols are exposed by the server.
+The last three are APIs that can be invoked programmatically. Every compiled API model is served through GraphQL.
+The `protocols` [compiler configuration](configuration.md#compiler-compiler) controls whether generated operations are exposed through REST and MCP.
 
 ## Data Products
 
@@ -49,7 +50,12 @@ Hidden columns, i.e. columns where the name starts with an underscore `_`, are n
 
 #### Schema Generation
 
-The compiler generates the GraphQL schema automatically from the SQRL script if no API schema is defined in the [`package.json`](configuration). Add the `--api graphql` flag to the [`compile` command](compiler.md#compile-command) to write the schema to the `schema.v1.graphqls` file in the same directory for inspection or fine-tuning.
+If no GraphQL schema is configured, the compiler infers one from the SQRL script and writes it to `build/inferred_schema.graphqls`.
+
+To provide and customize a schema for the default `v1` API, configure its path as `script.graphql` in [`package.json`](configuration.md#source-files-script).
+To serve multiple API versions, configure each version and its required schema under `script.api`. See [API Versioning](configuration.md#api-versioning).
+
+A useful customization workflow is to compile without a configured schema, copy `build/inferred_schema.graphqls` into the project (for example, `api/schema.v1.graphqls`), then configure that file as the API schema.
 
 #### Schema Customization
 
@@ -125,7 +131,8 @@ Selecting event times and totals together costs a single combined aggregate quer
 
 #### Authoritative Model
 
-DataSQRL uses the GraphQL schemas the authoritative model for all API protocols. It serves as the foundational model on which operations, endpoints, and access patterns are defined. This simplifies the conceptual model and server execution since any API operation maps to a GraphQL query which is executed by a centralized and optimized GraphQL engine.
+DataSQRL uses the GraphQL schema as the authoritative model for all API protocols. It is the foundation for operations, endpoints, and access patterns.
+This simplifies the conceptual model and server execution since any API operation maps to a GraphQL query which is executed by a centralized and optimized GraphQL engine.
 
 The GraphQL query execution engine sits at the core of the DataSQRL server engine and executes all requests even if the GraphQL API is not exposed. This ensures uniform execution of all requests and a shared authentication and authorization mechanism for security.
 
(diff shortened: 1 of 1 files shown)


documentation/docs/interface.md
```diff
@@ -9,7 +9,8 @@ Based on the SQRL script, DataSQRL generates the interface for the compiled data
 
 For data products, DataSQRL generates view definitions as deployment assets in `build/deploy/plan` which can be queried directly.
 
-The last three are APIs that can be invoked programmatically. The `protocols` [compiler configuration](configuration.md#compiler-compiler) controls which API protocols are exposed by the server.
+The last three are APIs that can be invoked programmatically. Every compiled API model is served through GraphQL.
+The `protocols` [compiler configuration](configuration.md#compiler-compiler) controls whether generated operations are exposed through REST and MCP.
```
mbroecheler, line 13: Don't we also support ONLY exposing REST or MCP? Let's say I only want to enable the REST API. The GraphQL model remains authoritative. But if I define explicit operations and use OPS_ONLY, I only get those exact operations exposed as REST endpoints, right? Or is there always a GraphQL API. If so, why?

documentation/docs/interface.md
```diff
@@ -185,7 +192,38 @@ This defines an operation `GetPersonByAge` which is the name of the MCP tool and
...
 The doc strings for the operations are used in the API and tooling documentation.
 
-By default, DataSQRL will add the custom operations to the generated ones. To only expose explicitly defined operations set `endpoints` option to `OPS_ONLY` in the [`package.json`](configuration).
+By default, DataSQRL adds explicit operations to the generated ones. Set `compiler.api.endpoints` to `OPS_ONLY` in the [`package.json`](configuration) to omit generated MCP and REST operations. The GraphQL endpoint remains available.
+
+### OpenAPI
+
+OpenAPI describes the REST API derived from GraphQL operations. It is not a second API definition.
+For every compiled API version, DataSQRL generates an OpenAPI 3 document from operations that expose a REST endpoint.
+The document includes REST paths, parameters, request bodies, and response schemas, but does not describe the GraphQL or MCP endpoints.
```
mbroecheler: replace `but does not describe the GraphQL or MCP endpoints.` by `for the REST endpoint only`.

documentation/docs/interface.md
```diff
@@ -185,7 +192,38 @@ This defines an operation `GetPersonByAge` which is the name of the MCP tool and
...
+The same configuration can customize the documentation endpoints and metadata, including `endpoint`, `uiEndpoint`, `title`, `description`, `version`, contact details, and license details.
+Compilation still generates the specification artifact.
+
+### GraphQL and OpenAPI Configuration
+
+GraphQL defines each API version, and OpenAPI is generated from that version's REST operations.
+The configuration determines which GraphQL schema is used and whether compilation checks the generated OpenAPI specification for backward compatibility.
+
+The `script.api.<version>.openapi` field points to a previously generated OpenAPI document for that compatibility check.
+It is not used as the served specification and does not define or customize the API.
```
mbroecheler, line 215: add:
To customize the API, you update the GraphQL schema or GraphQL operations that define the endpoints as described above.

documentation/docs/interface.md
```diff
@@ -185,7 +192,38 @@ This defines an operation `GetPersonByAge` which is the name of the MCP tool and
...
+| Neither `script.graphql` nor `script.api`        | Inferred as `v1`            | A `v1` OpenAPI artifact is generated; no compatibility check runs.                                                     |
+| `script.graphql`                                 | Configured as `v1`          | A `v1` OpenAPI artifact is generated; no compatibility check runs.                                                     |
+| `script.api.<version>.schema`                    | Configured for each version | An OpenAPI artifact is generated for each version.                                                                     |
+| `script.api.<version>.schema` and `.openapi`     | Configured for each version | The generated specification is compared with the configured prior document; compilation fails on incompatible changes. |
+| `script.api.<version>.openapi` without `.schema` | None                        | Invalid configuration: every `script.api` version requires `schema`.                                                   |
```
mbroecheler, line 225: Why does this not work against the generated GraphQL schema?
If I'm given a OpenAPI spec to implement (and assuming it maps correctly), it would be nice if I could configure it here and only write the SQRL code so that it maps correctly via the inferred graphql schema. Having to maintain a graphql schema in the middle that i don't need seems like overhead.

mbroecheler (changes requested):
Thanks for writing this up. I added some discussion points.

documentation/docs/interface.md
ferenc-csaky: In the current impl GraphQL is always publicly exposed. REST and MCP bridges hold a reference to `GraphQLServerVerticle` and execute their generated GraphQL operations directly through its GraphQL engine. That verticle currently both creates the engine and registers the `/v1/graphql` HTTP/WebSocket route, so deploying REST or MCP also deploys the public GraphQL route.

With `compiler.api.protocols: ["REST"]` and endpoints: `OPS_ONLY`, the generated bridge surface can be REST-only and restricted to the explicit operations. The same applies to MCP. But this does not currently disable or restrict direct GraphQL queries against the full configured schema.

documentation/docs/interface.md
ferenc-csaky: You’re right about this should be supported. The OpenAPI compatibility comparison itself does not require a configured GraphQL schema, as it runs after we generate the GraphQL model.

Although this now has a config limitation ATM: `openapi` is only accepted under `script.api.<version>`, which currently requires schema, and schema inference only runs when no API configuration is present. As a result, currently there is no way to provide an OpenAPI compatibility baseline for the inferred `v1` schema.

I can address this in a follow-up PR. Do we wanna support `openapi` under the non-versioned fields as well or is it enough to make `script.api.<version>` not require the schema? Or just the former? :)

mbroecheler (approved):
Thank you, let's move the follow on work in a separate PR

PR #2380, opened 2026-09-16
feat: Infer GraphQL schema for OpenAPI-only API versions

ferenc-csaky:
## Key changes
  - Allow versioned `script.api.<version>` configurations to omit `schema` when `openapi` is configured.
  - Infer the GraphQL schema from the SQRL script for OpenAPI-only API versions.
  - Preserve each API version’s configured version and GraphQL operations.
  - Require every versioned API configuration to define either `schema` or `openapi`.
  - Add documentation and tests for valid OpenAPI-only and invalid schema-less API configurations.


mbroecheler:
`Require every versioned API configuration to define either schema or openapi.`

How about this scenario: I want to use the inferred graphql schema but define explicit operations against it for the REST/MCP APIs (no openapi schema)?

mateczagany (approved):

PR #2381, opened 2026-09-17
feat(api): Support protocol-specific API exposure

ferenc-csaky:
## Key changes

  - Derive server GraphQL exposure and `OPS_ONLY` behavior from `compiler.api.protocols` and `compiler.api.endpoints`.
  - Allow REST-only, MCP-only, and REST/MCP-only deployments without registering public GraphQL HTTP, WebSocket, or GraphiQL routes, while retaining the internal GraphQL engine for bridge execution.
  - Enforce `OPS_ONLY` by accepting only explicitly configured named operations and executing their stored GraphQL definitions.
  - Return descriptive errors for unknown operations and GraphQL `InvalidSyntax` errors for malformed request documents.
  - Refactor server configuration conversion into extensible `ServerConfigConverter` implementations.
  - Document protocol and endpoint behavior, and add unit and integration coverage for REST-only and operation-only configurations.


ferenc-csaky:
I just added this as a starter, because I took a deeper look to see what else we may miss, and there are a bunch of things:

  - **High — No REST/MCP test transport:** `DatasqrlTest` only sends raw GraphQL documents to the GraphQL endpoint. REST-only and MCP-only deployments therefore cannot be tested end to end through the framework.
  - **High — No GraphQL request envelope:** Its raw `application/graphql` transport cannot express `variables`, `operationName`, extensions, or persisted-query metadata. Parameterized configured operations cannot be successfully tested.
  - **High — Generated queries conflict with `OPS_ONLY`:** The test planner automatically adds schema-derived GraphQL queries. Under `OPS_ONLY`, those queries are intentionally unconfigured and rejected, so the framework cannot cleanly run only explicitly configured operations.
  - **Medium — Global API configuration:** `compiler.api.protocols` and `compiler.api.endpoints` cannot vary by `script.api.<version>`, so all deployed API versions share the same public protocols and endpoint strategy.
  - **Medium — No version-specific test assignment:** Every `.graphql` test file runs against every API version; tests cannot target one version.
  - **Low — Incomplete multi-operation document support:** Named fragments are omitted when operations are sent individually, and multiple operations in one file share the filename-derived snapshot name.

cc @mbroecheler

mbroecheler:
Response to comment above:

- High — No REST/MCP test transport: DatasqrlTest only sends raw GraphQL documents to the GraphQL endpoint. REST-only and MCP-only deployments therefore cannot be tested end to end through the framework.
  - True, for testing we would have to overwrite the protocols and make sure GraphQL is always present. So, in that sense it is a bit weird that you are developing a REST-only API and then have to write your queries in GraphQL. We could look into reusing the defined operations so you don't have to copy that into the tests. But I don't see that as a huge issue because REST and MCP are straight mappings from GraphQL that we can validate at compile time (i.e. we can ensure that the defined operations are valid GraphQL queries) and properly exercise through GraphQL. The only thing we are missing is actually exercising the REST and MCP mapping routes, but that piece is not dependent on a particular project implementation, so any bugs there should be found by our SQRL testing harness. If the user wishes, they can do a `cmd run` combined with a python script to test REST if it makes them feel better. But it should not provide any extra "correctness" or "regression" gurantees that GraphQL alone does not.
- High — No GraphQL request envelope: Its raw application/graphql transport cannot express variables, operationName, extensions, or persisted-query metadata. Parameterized configured operations cannot be successfully tested.
  - Is this for the testing framework? I can run queries through GraphiQL with variables without problem (last time I checked). Same comment as above applies: these exercise features at the framework level, not at the individual implementation level. 
- High — Generated queries conflict with OPS_ONLY: The test planner automatically adds schema-derived GraphQL queries. Under OPS_ONLY, those queries are intentionally unconfigured and rejected, so the framework cannot cleanly run only explicitly configured operations.
  - This is for tests defined via hint in SQRL, right? We would have to relax this for testing.
- Medium — Global API configuration: compiler.api.protocols and compiler.api.endpoints cannot vary by script.api.<version>, so all deployed API versions share the same public protocols and endpoint strategy.
  - I don't see this as a problem. I cannot really think of a use case where you want some protocols in version 1 and others in version 2. It may organically grows (i.e. we add a REST endpoint) but it wouldn't be a problem if version 1 then also gets a REST endpoint retroactively.
- Medium — No version-specific test assignment: Every .graphql test file runs against every API version; tests cannot target one version.
   - This is something we should probably support via the `.properties` files that we use for the auth header. Add a field `api-versions` in there which accepts a single or array value and only executes against specified versions or all if blank. Not a high priority item, but will be necessary eventually. That said, do we fully support multiple versions now in the implementation?
- Low — Incomplete multi-operation document support: Named fragments are omitted when operations are sent individually, and multiple operations in one file share the filename-derived snapshot name.
   - This is also for testing only, right? Good limitation to keep in mind and file a ticket for. Not sure it's a priority to fix.


BTW, these are all my personal opinions. Users might feel differently - we will find out. But I don't see a need for urgent action on any of these items. The most important follow up is checking to what extend we actually support multiple API versions ;-)

mateczagany (approved):

ferenc-csaky:
> Is this for the testing framework? I can run queries through GraphiQL with variables without problem (last time I checked). Same comment as above applies: these exercise features at the framework level, not at the individual implementation level.

> This is for tests defined via hint in SQRL, right? We would have to relax this for testing.

Yes, both are testing-framework limitations, not runtime API limitations.

